Issue #25, opened 2022-11-11
$AMPE balance not visible


prayagd:

erasmus:
@prayagd can you add the steps to reproduce the bug + console log with any errors?

nicolasdanelon:
@gonzamontiel @ebma I've fix this bug, here's the PRs for master and dev https://github.com/pendulum-chain/portal/pulls
cc @prayagd @erasmus

PR #30, opened 2022-11-23
Bugfix/ampe balance not visible (master)

nicolasdanelon:
fix https://github.com/pendulum-chain/portal/issues/25

### DIFF
--- a/src/pages/dashboard/Dashboard.tsx
+++ b/src/pages/dashboard/Dashboard.tsx
@@ -28,7 +28,7 @@ export function Dashboard() {
     feeFrozen: BigInt(0),
   });
 
-  const { feeFrozen, miscFrozen, free } = accountBalance;
+  const { free } = accountBalance;
 
   if (free) {
     console.log("Free proc: ", toUnit(free) / 1000);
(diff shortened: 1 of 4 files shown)


prayagd:
@nicolasdanelon I confirm i can see the AMPE balance.

src/pages/dashboard/Dashboard.tsx
```diff
@@ -37,32 +37,23 @@ export function Dashboard() {
...
-  };
-
   return (
     <div class="mt-10">
       <div className="dashboard portfolio">
         <h1>Portfolio</h1>
         <div className="portfolio">
           <h4>Total balance</h4>
-          <h2>{maxBalance()} AMPE</h2>
+          <h2>{toUnit(free) / 1000} AMPE</h2>
```
ebma, line 56: I think I would prefer if this value was wrapped in a `useMemo()` hook to prevent unnecessary re-renders of the component.

ebma (commented):
LGTM overall but I didn't check if the balance works because I don't have an account with AMPE.

src/pages/dashboard/Dashboard.tsx
gonzamontiel: Agreed with this comment ^ 
Also dividing by 1000 here doesn't seem a good idea (maybe include it in the toUnit()

gonzamontiel (changes requested):

gonzamontiel (approved):

PR #34, opened 2022-12-01
Adding foucoco

nicolasdanelon:


gonzamontiel (dismissed):
It's not very clear to me when Foucoco is used (whether the user can switch between networs or not), but I'll approve anyway, the rest is fine.

EDIT: redirection seems not to work properly, please review:
When entering:
http://localhost:5173/foucoco/dashboard -> http://localhost:5173/amplitude/dashboard
http://localhost:5173/foucoco/collators -> http://localhost:5173/amplitude/collators

It only works when typing:
http://localhost:5173/foucoco
http://localhost:5173/amplitude

PS: a switch in the UI would be appreciated

gonzamontiel (changes requested):
see my comment above

gonzamontiel (approved):
I fixed the redirection

PR #44, opened 2023-01-23
37 update pendulum branding

gonzamontiel:
Draft PR for pendulun branding, merges to `develop`


gonzamontiel:
I additionally tried to fix the collators table, following the designs, it's not perfect but since it was not part of the ticket I think is decent.

ebma:
It's looking really nice, good job. Two things that I noticed when playing around with it:
1) the text color of for example this button is not ideal
image

and 2), the text in the active tab is white on white, making it unreadable.
image


We can tackle these problems later but maybe fixing them here is a good idea because I assume it can be done in the root CSS file (?)

gonzamontiel:
> It's looking really nice, good job. Two things that I noticed when playing around with it:
> 
> 1. the text color of for example this button is not ideal
>    image
> 
> and 2), the text in the active tab is white on white, making it unreadable. image
> 
> We can tackle these problems later but maybe fixing them here is a good idea because I assume it can be done in the root CSS file (?)

Hey, thanks for the review. You are right about the colors but, I didn't even look at those tabs. I will take a look and see if it's worth a little change, but honestly those pages will look bad unless I tweak then completely

ebma:
Okay, you're right it might not make sense to tweak these things if we don't need the AMM tab anyways for now, and we have to change the other things (like bridging tabs and collator dashboard) again anyways. So I don't mind if you don't spend more time on this now.

ebma (approved):
I added another eslint plugin and updated the linting script to also consider .tsx files. Having eslint tell us about eg missing dependencies of hooks is amazing. Thanks for also integrating that in the PR. 

Ready to merge from my side.

PR #47, opened 2023-02-04
Redeem tokens on stellar network

### LINKED ISSUE #40: Redeem tokens on Stellar network - Spacewalk
As a wrapped stable coin holder on Amplitude network, I want to bridge my tokens back to Stellar.

# Acceptance criteria
* On `\bridge`
* At top of the page show warning
   * Currently in beta mode. Transactions are capped to $9999. 
* Show `Issue` and `Redeem` tab. 
* In `Redeem`. Show Input field and a dropdown next to it
   * Input field to enter the amount of tokens to be bridged.
   * Dropdown to select the token tickers.
* Below this show one more input field. To enter Stellar address. 
* Next show the amount of tokens and the token ticker to be received on Stellar
* At the bottom show `Confirm` button. On click opens-up a pop-up.
* In pop-up
   * Show tokens redeemed and their value in dollars.
   * Show the stellar address the tokens are sent to.
   * At the bottom of pop-up show button `View Progress`
* On clicking `View Progress` pop-up closes and directs you the particular transfer on `\transfer`

### Notes
* Following ticket is written based on assumptions that vaults and fee structure are not defined. 
* Input field - Inputs only until two decimals allowed. 

### Wireframe
Will update once ready

# Technical consideration
To be filled in by the tech-team.

ebma:
Closes #40.

Unfortunately, it shows lots of commits in the history. It's probably because I based it off the branch for issuing, which was squash-merged into staging.


ebma:
@gonzamontiel this is ready for review now.

src/helpers/parseNumbers.ts
```diff
@@ -10,17 +17,37 @@ BigNumber.NE = -20;
 // Converts a decimal number to the native representation (a large integer)
 export const decimalToNative = (value: BigNumber | number | string) => {
   const bigIntValue = new BigNumber(value);
-  const multiplier = new BigNumber(10).pow(TOKEN_DECIMALS);
+  const multiplier = new BigNumber(10).pow(AmplitudeDecimals);
```
gonzamontiel, line 20: Okay with the lowercase buuut, is it only amplitude? I think we have 12 decimals in all our chains.

src/helpers/parseNumbers.ts
```diff
@@ -10,17 +17,37 @@ BigNumber.NE = -20;
...
+};
+
+export const decimalToStellarNative = (value: BigNumber | number | string) => {
+  const bigIntValue = new BigNumber(value);
+  const multiplier = new BigNumber(10).pow(StellarDecimals);
 
   return bigIntValue.times(multiplier);
 };
 
+export const fixedPointToDecimal = (value: BigNumber | number | string) => {
```
gonzamontiel, line 32: why is this method called from fixedPoint?  It receives the same possible values than `decimalTo...` methods, and it treats  the value as a BigInt too. What it should receive then?

gonzamontiel (approved):
looks good to me! I didn't test it, though. Only left some comments.

src/helpers/parseNumbers.ts
ebma: There are now three different things we need to distinguish about numbers.
Our fee parameters (or most things that represent a percentage in spacewalk pallets) are represented by an UnsignedFixedPoint with `u128` as the inner type. This means that `1` or `100%` is represented by `1,000,000,000,000,000,000` ie 1e18, `10%` would be `100,000,000,000,000,000` etc. 
When it comes to balances, we need to consider the `decimals` parameter that the chain is using. On our chains that's `12` for each network. 
Lastly, Stellar assets only have a precision of 1e7. Thus we need a third conversion function for that.

It can be a bit confusing of course, but I think having three different helper functions with appropriate names helps.

src/helpers/parseNumbers.ts
ebma: True. We could rename it to `ChainDecimals` maybe and neglect the fact that this could be different for each chain until we actually encounter the case where we need to use different decimals (which is unlikely to happen anytime soon).

ebma:
I did not find the time to clean this up yet, sorry for that. @gonzamontiel would you mind cleaning this up ie resolving the merge conflicts so we can merge it soon? Feel free to add some changes about your previous comments of course

gonzamontiel:
> 

Sure thing, leave it on me. I will address the conflicts and my comments.

src/helpers/parseNumbers.ts
gonzamontiel: Okay, thanks a lot to clarify. I will then leave it like that then.

ebma:
The changes look good to me but can we do something about the CI failing? I assume it's mainly because of the react-table thing being hard to type properly. Since it's actually not related to this PR but to the collators list maybe it's okay if we don't fix it here.

PR #57, opened 2023-02-21
feat: swap

nejcm:


src/constants/cache.ts
```diff
@@ -0,0 +1,49 @@
+import { UseQueryOptions } from '@tanstack/react-query';
+
+export const cacheKeys = {
```
gonzamontiel, line 3: can you elaborate a bit more on how this cache would work?

gonzamontiel (changes requested):
I like most of the changes introduced :)
Disclaimer: I reviewed considering that changes are mostly UI; data fetching and full behavior is still missing. For that reason I skipped some TODOs and other minor details.

There are quite a lot of changes here, and not only related to amber, but I think most of them are positive and organize better the code, like the one in the Nav links, so thanks for that!

Some remarks:
1. Sidebar shadows and box shadows do not follow the figma designs
2.  Amplitude theme for swap looks a bit broken (maybe it's just one variable that repeats over different places)
Screenshot from 2023-04-18 13 50 57

3. I noticed you introduced Yup as an schema validator, I'm not familiar with it but looks good. Maybe we should adopt it for the whole app, delegation and spacewalk dialogs, wdyt @ebma ?
4. I noticed you introduced some kind of cache (from react-query I think), why was that ? I am not against relying on external libraries, but they come with the extra cost of loading time and performance. But if this is important for usability let's leave it.

nejcm:
> I like most of the changes introduced :) Disclaimer: I reviewed considering that changes are mostly UI; data fetching and full behavior is still missing. For that reason I skipped some TODOs and other minor details.
> 
> There are quite a lot of changes here, and not only related to amber, but I think most of them are positive and organize better the code, like the one in the Nav links, so thanks for that!
> 
> Some remarks:
> 
> 1. Sidebar shadows and box shadows do not follow the figma designs
> 2. Amplitude theme for swap looks a bit broken (maybe it's just one variable that repeats over different places)
>    Screenshot from 2023-04-18 13 50 57
> 3. I noticed you introduced Yup as an schema validator, I'm not familiar with it but looks good. Maybe we should adopt it for the whole app, delegation and spacewalk dialogs, wdyt @ebma ?
> 4. I noticed you introduced some kind of cache (from react-query I think), why was that ? I am not against relying on external libraries, but they come with the extra cost of loading time and performance. But if this is important for usability let's leave it.

Yes there are some small fixes and changes to existing UI not connected to amber to improve how sidebar behaves on different screens. I will address you remarks and fix them. But yes this PR contains only UI and stubs for future data fetching.

1. Will fix
2. So I worked out of figma and currently only the light theme design is there. So work will be needed to address the amplitude dark theme. 
3. Yup is very popular and it is very easy to integrate with hook-form that you guys already use. It also simplifies the validation of forms.
4. So the react-query is an asynchronous state management library (mostly used for fetching data) (https://tanstack.com/query/latest) that offers many things out of the box like: loading states, caching, pagination, infinite queries, refetching and many other cool functionality. Now you might replace it with apollo graphql one in case fetching from indexers with be using graphql queries but that will be quite easy as they have a very similar usage signature. So currently I am using it mostly cause it makes it very easy to handle loading states and also success or error handling.

nejcm:
And I see now the figma UI was updated from when I last checked. Will reflect the changes.

gonzamontiel (approved):
thanks! lets get this merged :)

gonzamontiel:
> 2. So I worked out of figma and currently only the light theme design is there. So work will be needed to address the amplitude dark theme.

I didn't know that (cc @prayagd), but out of assumptions, we have a basic system for background colors that you can rely on. Like `bg-base-100` and `bg-base-200` should work for both themes out of the box, same with some defined variables like `--table-text`, etc.

nejcm:
@gonzamontiel Thanks. As discussed once we start working on the functionality part I will split up the PRs based on the open tasks.

PR #63, opened 2023-02-28
feat: useful hooks

nejcm:
As mentioned this includes the local storage hook in case it might be useful.


gonzamontiel (changes requested):
I am not sure about the useBoolean hook, is it required? If not, can you pls remove it?
The rest looks good, we can merge it and then incorporate the hook in some other PR.

nejcm:
> I am not sure about the useBoolean hook, is it required? If not, can you pls remove it? The rest looks good, we can merge it and then incorporate the hook in some other PR.

Not needed with this PR but I am using it on the swap PR as I am controlling some state of dropdowns and modals with it.

gonzamontiel (approved):

Issue #62, opened 2023-02-28
Add pagination to Collator's list


gonzamontiel:
Collators table should include pagination, as shown in the designs:

![Screenshot from 2023-02-28 10-08-31](https://user-images.githubusercontent.com/4713407/221805719-6cdcc8af-348c-42ed-9d21-d6c2fb3fbfc4.png)

### Technical considerations
- There's a related PR that includes a paginated Table (see #64).
- Replace the `<table>` component in `Collators.tsx` with  the paginated component

nejcm:
Will open another PR for this and the pagination one can be merged already

PR #64, opened 2023-02-28
feat: pagination #62

nejcm:
@gonzamontiel Let me know if this satisfies your needs or is there something more needed. It is as designed on the screenshot. This is only a reusable component so it can be included anywhere not just tables.
It can also be extend to support actual page numbers or even a page selector.


gonzamontiel:
So this component is necessary indeed, thanks for that. But what I had in mind for #62 was something like `PaginatedTable`, that includes the pagination section _and_ the behavior to load the rows when clicking on the chevron icons. 
So, I can approve this PR as a previous step, but #62 would still be open.

gonzamontiel (approved):
As I stated in my comment, a good previous step. lgtm!

nejcm:
> So this component is necessary indeed, thanks for that. But what I had in mind for #62 was something like `PaginatedTable`, that includes the pagination section _and_ the behavior to load the rows when clicking on the chevron icons. So, I can approve this PR as a previous step, but #62 would still be open.

I see. Will make a new PR with that today.

nejcm:
@gonzamontiel I updated this PR with the reusable table but it uses the react-table library to display it. You can check [here](https://github.com/pendulum-chain/portal/pull/65/files#diff-193ebe625f046c7606d7806db8425ebf852aefed98ea91762459f5259fa70132) how to  define columns for it which is the only required part for using it.

nejcm:
@gonzamontiel Not sure what is your merge policy but will let you merge this in once you think its ok

gonzamontiel (approved):
It also comes with a global search? nicee. Looks good to me!

gonzamontiel:
Oh, I see it has some conflicts, feel free to merge it when they are solved @nejcm :)

nejcm:
@gonzamontiel Solve the merge issues. I dont have the perms to merge though

PR #70, opened 2023-03-08
feat: Collators table

### LINKED ISSUE #62: Add pagination to Collator's list

nejcm:
Closes #62

### DIFF
--- a/src/pages/collators/Collators.tsx
+++ b/src/pages/collators/Collators.tsx
@@ -185,6 +185,7 @@ export function Collators() {
         columns={columns}
         isLoading={!candidates}
         search={false}
+        pageSize={8}
       />
     </div>
   );
--- a/src/pages/collators/Collators.tsx
+++ b/src/pages/collators/Collators.tsx
@@ -1,44 +1,36 @@
-/* eslint-disable react/prop-types */
-/* eslint-disable react/jsx-key */
-// @ts-nocheck
 /**
  * FIXME remove @ts-nocheck, it was specifically added because of some errors in react-table.
  * Probably fixed in https://github.com/pendulum-chain/portal/pull/64
-*/
-import { Button } from "react-daisyui";
-import { useEffect, useMemo, useState } from "preact/hooks";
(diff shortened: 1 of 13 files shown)


src/pages/collators/Collators.tsx
```diff
@@ -50,133 +42,76 @@ export function Collators() {
...
           collator: candidate.id,
           totalStaked: format(totalStaked, tokenSymbol),
           delegators: candidate.delegators.length,
-          apy: inflationInfo?.delegator.rewardRate.annual || "0.00%",
+          apy: inflationInfo?.delegator.rewardRate.annual || '0.00%',
         };
-        return rowItem;
       }),
-    [candidates, inflationInfo?.delegator.rewardRate.annual, tokenSymbol]
+    [candidates, inflationInfo?.delegator.rewardRate.annual, tokenSymbol],
```
gonzamontiel, line 97: This extra comma is from the lint? is it necessary?

gonzamontiel (commented):
As we spoke, the idea of the pagination was to avoid the overflow that occurs here.
An easy fix (instead of calculating rows height and viewport height), would be to calculate, in average the amount of rows that should fit in the viewport in a Desktop, and that will work in most cases. 
Also, didn't find the culprit, but the table border shouldn't be that clear, but like the inner rows border.

Screenshot from 2023-03-14 14 49 06

nejcm:
@gonzamontiel Updated

gonzamontiel (approved):
I've added some changes and I think it is ready to merge.
The advantage of having an abstract Table with pagination (thanks @nejcm) comes with an inevitable loss in code readability.
We don't have the time now to make it perfect, but I think there is room for improvement.
We should keep in mind for future PRs

Issue #61, opened 2023-02-27
State is not saved after reloads


gonzamontiel:
After connecting a wallet to the portal, and reloading the page, the wallet is automatically disconnected.

1. Connect a wallet
2. Check that your balance is something
3. Reload the page
4. Check that your wallet is not connected anymore and shows no balance or other functionality related to it.

The expected behavior would be to have the wallet persistent after reloading the page.

Proposed solution:
We should save the global state to (and load it from) local storage. We should also add a way of disconnecting the wallet.

nejcm:
I have a useful hook for local storage I am already using in swap component for persisting the timeout and tolerance. Will make a separate PR for it so you can use it if needed.

gonzamontiel:
Hey @nejcm , I have assigned this one to you from what we spoke earlier, hope that is okay :)

PR #71, opened 2023-03-08
feat: save global state to storage

nejcm:
A couple of notes on this: 
- needs to be tested from your side as state handling was a bit refactored
- an improvement would be to have an expiration date also
- I am not sure about security implications of this as we are storing the wallet address in storage. Now it is true that this is specific to the browser of the user where also the wallet is used/installed but would still like to do some research on this before going to prod with this.
- note that the wallet info get stored in storage and the address is shown in the button but than the ss58Format comes in afterwards so the address changes in the button after 2-4s and that formatted one is not stored. Not sure how to fix this flicker


nejcm:
Putting this into draft. Discovered a small issue that needs addressing

gonzamontiel:
Hey @nejcm , thanks for your PR. 

> A couple of notes on this:
> 
> * needs to be tested from your side as state handling was a bit refactored
> * an improvement would be to have an expiration date also
I thought based on the implementation in the [other PR](https://github.com/pendulum-chain/portal/pull/63) that this timeout mechanism was already there, it isn't the case?

> * I am not sure about security implications of this as we are storing the wallet address in storage. Now it is true that this is specific to the browser of the user where also the wallet is used/installed but would still like to do some research on this before going to prod with this.

Thanks for bringing this up. I believe we don't store any secret key for the user, since the accounts are managed by an extension (Talisman or PolkadotJS). These extensions ask for a password to securely store the address. How secure this is? Probably not ideal, but we are not less safe than the wallets we use.

> * note that the wallet info get stored in storage and the address is shown in the button but than the ss58Format comes in afterwards so the address changes in the button after 2-4s and that formatted one is not stored. Not sure how to fix this flicker
I will test this and come back to you, 2-4s is quite a lot.

erasmus:
@nejcm can I ask you to please connect your PRs with the issue? 🙏

nejcm:
it is connected. https://github.com/pendulum-chain/portal/issues/61

nejcm:
Merged staging

src/pages/collators/dialogs/ClaimRewardsDialog.tsx
```diff
@@ -65,70 +60,74 @@ function ClaimRewardsDialog(props: Props) {
...
+    userRewardsBalance,
+  ]);
 
   const content = useMemo(() => {
     switch (step) {
       case ClaimStep.Confirm:
         return (
           <div className="rounded-lg bg-base-200 flex flex-col p-8 items-center w-fit center m-auto">
             <p className="flex">Amount</p>
-            <h1 className="flex text-4xl">{format(amount, tokenSymbol, true)}</h1>
```
gonzamontiel, line 97: Hey @nejcm, do you have an auto formatter ? In that cases, which one? So we all agree in one with the same rules.
This causes that every single file is changed to a new format, leading to conflicts with lines that you didn't even touch. I can't approve this PR with so many unrelated lines changed

gonzamontiel (changes requested):
So many unrelated lines are changed due to formatting. We better first agree on format to avodi conflicts.

nejcm:
I have prettier which is included in the project with the relevant config. The only difference is that I have a config option set that it runs each time I save the file. So each file I change it will run it on the entire file. I believe the prettier was not run on the entire project yet since its inclusion.

gonzamontiel:
hey @nejcm , there are some conflicts due to recent changes on staging, hopefully after those everything will be smoother. Can you please fix them and then I'll approve! thanks :pray:

src/components/LabelledInputField/index.tsx.orig
```diff
@@ -0,0 +1,77 @@
...
+  step?: string;
+  value: string;
+  extraBtnText?: string;
+  extraBtnAction?: () => void;
+  onChange?: (value: string) => void;
+  style?: React.CSSProperties;
+  errorMessage?: string;
+}
+
+<<<<<<< HEAD
```
gonzamontiel, line 21: Also another unresolved conflict.

gonzamontiel (changes requested):
hey @nejcm pls check that there are unresolved conflicts. Better would be to rebase the branch to staging.

src/assets/UnlinkIcon.tsx.orig
```diff
@@ -0,0 +1,23 @@
+interface Props {
```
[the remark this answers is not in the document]
nejcm: Missed it and did not delete the orig file

gonzamontiel:
Hey, there are quite a lot of orig files. Maybe it is better to create another, simple PR, with only the requested changes (save state to local storage). It is too hard to review at this point.

nejcm:
@gonzamontiel I believe all removed now

gonzamontiel (approved):

PR #78, opened 2023-03-09
27 add undelegation functionality

### LINKED ISSUE #27: Add undelegation functionality for collators
As $AMPE holder, I want to undelegate my tokens from a collator.

# Acceptance criteria

### Undelegate tokens
* On `\collators`
* In the `Collators` container show
    * Unlock amount
    * Unlock time
* Show 'Undelegate' button next to delegated amount. Only show this button when tokens are delegated
* Clicking on this `Undelegate` opens a pop-up
* In pop-up show:
    * Address of the collator
    * Delegated amount
    * Show Input field to enter the amount of tokens.
        * If input value greater than delegated amount amount show error. Error message `Amount Incorrect`
        * If input value has more than two decimals show error. Error message `Incorrect input. Only two decimals`
        * Show all error messages below input field in red color. 
* Show `Undelegate` button. On click.
    * Initiate the on-chain transaction using connected wallet.  
        * On transaction failed
            * Show "Failed" pop-up
        * On transaction successful
            * if Input amount == Delegation amount
                * On transaction successful
                    * Show "Successful" pop-up
                    * Enable `Delegate` buttons of all collators and disable the `Undelegate` from this specific collator.
                    * Amount delegated will be unlocked after 7 days from the time of transaction execution
                    * Update the unlock amount in the top container of the dashboard with input amount.
                    * Update the unlock time in the top container to 7 days from the time of transaction and start a countdown. 
                    
### Terms used in the wireframe
* Available balance: User’s `AMPE Balance`
* Delegated amount: `The amount delegated by the user to that collator`

### Notes
* All data is read only. Except buttons and input field. Which are explicitly mentioned.

### Wireframe
https://www.figma.com/file/XLIKffEYnK9autHMM1YgA3/Flows?node-id=0%3A1&t=RWBKfzuRcngaIfGY-0

# Technical considerations
Most of the technical consideration in [issue 26](https://github.com/pendulum-chain/portal/issues/26) is also relevant for this issue.

Additionally:
* request information
  * determine whether the user is currently already delegating to any of the collators
    * `api.query.parachainStaking.delegatorState(<user account address>)`
    * this returns an object `value`, which contains a Boolean `value.isSome`
    * if `value.isSome` is `true`, then the user is currently delegating, if it is `false`, then the user is currently *not* delegating
  * the currently delegated amount and who the user is delegating to
    * if for the above value `isSome` is true, then you can call `value.unwrap()`
    * this will return an object with two members
       * `owner`: the address of the collator the user is delegating to, display nicely with `toHuman` (i.e., `value.unwrap().owner.toHuman()`)
       * `amount`: the delegation amount, be aware that this is the _raw_ amount again

gonzamontiel:
Closes #27 

Some things to consider:
- I reused the same dialogs than for delegation, changing texts and actions. Maybe a bit dirty (a couple of extra booleans and checks), but that won against duplicating everything.
- For that reason I also had to reset the booleans everytime onClose(), because the dialogs are always in the dom, and the state only manages their visibility.

### DIFF
--- a/src/pages/collators/Collators.tsx
+++ b/src/pages/collators/Collators.tsx
@@ -162,7 +162,6 @@ export function Collators() {
                   variant="outline"
                   onClick={() => {
                     setUnbonding(false);
-                    console.log("unbonding", unbonding);
                     setSelectedCandidate(row.original.candidate);
                   }}
                   disabled={!showDelegate}
--- a/src/pages/collators/Collators.tsx
+++ b/src/pages/collators/Collators.tsx
@@ -233,8 +233,10 @@ export function Collators() {
         userAvailableBalance={userAvailableBalance}
         userStake={userStaking?.amount}
         selectedCandidate={selectedCandidate}
-        isDelegatingMore={userStaking?.candidateId === selectedCandidate?.id}
-        isDelegatingLess={unbonding}
+        mode={unbonding
+          ? 'undelegating'
+          : userStaking?.candidateId === selectedCandidate?.id
+            ? 'delegatingMore' : 'joining'}
         onClose={() => {
           setSelectedCandidate(undefined)
           setUnbonding(false);
--- a/src/pages/collators/Collators.tsx
+++ b/src/pages/collators/Collators.tsx
@@ -26,6 +26,7 @@ import UnlinkIcon from "../../assets/UnlinkIcon";
 import ExecuteDelegationDialogs from "./dialogs/ExecuteDelegationDialogs";
 import ClaimRewardsDialog from "./dialogs/ClaimRewardsDialog";
 import { Balance } from "@polkadot/types/interfaces";
+import { fees } from "@polkadot/api-derive/contracts";
 
 interface CollatorColumnProps {
   candidate: ParachainStakingCandidate;
@@ -47,12 +48,13 @@ export function Collators() {
   const { candidates, inflationInfo, estimatedRewards } = useStakingPallet();
 
   // Holds the candidate for which the delegation modal is to be shown
-  const [selectedCandidateForDelegation, setSelectedCandidateForDelegation] =
+  const [selectedCandidate, setSelectedCandidate] =
     useState<ParachainStakingCandidate | undefined>(undefined);
 
   const [userAvailableBalance, setUserAvailableBalance] = useState<string>("0.00");
   const [userStaking, setUserStaking] = useState<UserStaking>();
   const [claimDialogOpen, setClaimDialogOpen] = useState<boolean>(false);
+  const [unbonding, setUnbonding] = useState<boolean>(false);
 
   const userAccountAddress = useMemo(() => {
     return walletAccount && ss58Format
@@ -145,7 +147,10 @@ export function Collators() {
                   className="mr-2 text-primary"
                   size="sm"
                   color="ghost"
-                  onClick={() => undefined}
+                  onClick={() => {
+                    setUnbonding(true);
+                    setSelectedCandidate(row.original.candidate);
+                  }}
                   startIcon={<UnlinkIcon className="w-4 h-4" />}
                   style={{ visibility: showUnbond ? "visible" : "hidden" }}
                 >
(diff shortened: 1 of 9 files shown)


src/pages/collators/Collators.tsx
```diff
@@ -47,12 +48,13 @@ export function Collators() {
   const { candidates, inflationInfo, estimatedRewards } = useStakingPallet();
 
   // Holds the candidate for which the delegation modal is to be shown
-  const [selectedCandidateForDelegation, setSelectedCandidateForDelegation] =
+  const [selectedCandidate, setSelectedCandidate] =
     useState<ParachainStakingCandidate | undefined>(undefined);
 
   const [userAvailableBalance, setUserAvailableBalance] = useState<string>("0.00");
   const [userStaking, setUserStaking] = useState<UserStaking>();
   const [claimDialogOpen, setClaimDialogOpen] = useState<boolean>(false);
```
ebma, line 56: Can we maybe add a comment here that this is basically a switch between bonding/unbonding mode? I get how and that it works but it's a bit implicit and in a few weeks getting back to this might be confusing. 
We could also change it to something like this
```
  const [bondingMode, setBondingMode] = useState<'bonding' | 'unbonding' | undefined>(undefined);
```

src/pages/collators/Collators.tsx
```diff
@@ -156,8 +161,9 @@ export function Collators() {
                   color="primary"
                   variant="outline"
                   onClick={() => {
-                    // eslint-disable-next-line react/prop-types
-                    setSelectedCandidateForDelegation(row.original.candidate);
+                    setUnbonding(false);
+                    console.log("unbonding", unbonding);
```
ebma, line 165: Remove the log maybe

src/pages/collators/dialogs/DelegateToCollatorDialog.tsx
```diff
@@ -19,6 +19,7 @@ interface DelegateToCollatorDialogProps {
   tokenSymbol: string;
   visible: boolean;
   isDelegatingMore: boolean;
+  isDelegatingLess: boolean;
```
ebma, line 22: We could also replace them with a `mode` so that we make it more explicit that they are mutually exclusive and are not supposed to be `true` both at the same time.

ebma: Similar to the `useState` I described before

src/pages/collators/dialogs/ExecuteDelegationDialogs.tsx
```diff
@@ -4,25 +4,34 @@ import {
...
+import { doSubmitExtrinsic } from "./helpers";
 
 interface ExecuteDelegationDialogsProps {
   userAvailableBalance: string;
-  selectedCandidateForDelegation: ParachainStakingCandidate | undefined;
-  isDelegatingMore: boolean;
+  userStake: string;
+  selectedCandidate: ParachainStakingCandidate | undefined;
+  isDelegatingMore?: boolean;
+  isDelegatingLess?: boolean;
```
ebma, line 21: Same here

ebma (commented):
I like the changes and that you found a way to make the existing dialogs reusable in both flows 👍 I left some comments with suggestions to improve code readability but there are no big issues I think

src/pages/collators/Collators.tsx
gonzamontiel: It is true that there's room for improvement here.
The way I see it, _selectedCandidate_ is selected in any case you interact with a candidate in a row. It's just the 'selected row' sort of, no matter what you do with that info later.

I am using isDelegatingMore and isDelegatingLess to manage the state internally in the dialogs. In that case, unbonding is implied by isDelegatingLess, and the other is implied by the negation of isDelegatingLess. It's its not perfect, I even considered replacing the booleans with a state like the one you suggest, that contemplates three states: bonding, bondingMore, unbonding, but I thought it was not worth it. Let me know your thoughs @ebma

src/pages/collators/dialogs/DelegateToCollatorDialog.tsx
gonzamontiel: Alright then, it'll be a bit more verbose but maybe more readable. Will fix it.

gonzamontiel:
Comments addressed. I think is ready to merge, wdyt @ebma ?

ebma (approved):
LGTM, thanks for changing it

PR #91, opened 2023-03-23
41 show list of spacewalk transfers

### LINKED ISSUE #41: Show list of spacewalk transfers
As a Stellar and Amplitude user, I should be able to track the status of my transfers on spacewalk. 

# Acceptance criteria
* On `\transfers`
* Show two categories `Issue Requests` and `Redeem Requests`.
* Show list of previous transfers under `Issue` and `Redeem` respectively.
* Show following information in the list. 
   * Date and time
   * Amount
   * Stellar transaction
   * Status (Completed, Pending and Failed)
* On clicking any of the list item. 
   * Show a new page with additional details of the transaction.
      * Parachain block
      * Destination address
      * Vault account
      * Stellar transaction ID (Only for issue requests)
      * Received amount of tokens (only if status completed)

### Wireframe
Will update once ready

# Technical consideration

* Use Spacewalk pallets storage to access the requests 
* Precisely, we need to call `api.query.issue.issueRequests.entries()` and `api.query.redeem.issueRequests.entries()` to retrieve such information
* Parse the information to extract the required data

gonzamontiel:
- I collected both Issue and Redeem requests and merged also the types, the resulting one would be: Pending' | 'Completed' | 'Cancelled' | 'Reimbursed' | 'Retried'
- I left Reimbursed and Retried with a white/gray color (I didn't have designs for them)
- The list is only accessible if typing `/<chain>/transfers` where <chain> is one of foucoco, amplitude, pendulum  
- I included an extra column for the bridged asset code
- I included an extra column to see the details (involving the whole row was inconvenient, and I think this solution works pretty well)
- I have formatted sometimes the numbers with the unit prefix like (mili, micro, millions). But in the transfer detail dialogs, I included the title property in the amount element, so the user can check the actual number with all the decimals. This functionality is subject to change if you think it's a bad UX.
- I based all the designs from the latest modifications in [figma flows](https://www.figma.com/file/XLIKffEYnK9autHMM1YgA3/Flows?node-id=192-2382&t=DGhNMcm28APkvQBB-0)
- **[IMPORTANT]** I have left unimplemented the reimburse and compensate buttons because it is not clear anywhere what to expect after they are clicked (the flow ends there). This is fine for me, since it can be implemented in another ticket. 

Closes #41


ebma:
@gonzamontiel good point. You would have to implement something similar to the [calculateDeadline](https://github.com/pendulum-chain/portal/blob/d92c3abe36dac9e40b416ea8d51b5051618f1ebd/src/helpers/spacewalk.ts#L76)() function.

The `opentime` is indeed not the block hash but it's the active block number that was the latest when the request was issued. Note that, the active block number is an extra counter that spacewalk has, which is only incremented if the oracle is running. If the oracle is offline because price data is not up-to-date, then this active block counter is not incremented per block. 
This is some detail that I would omit for simplicity, however. Thus, what you show as the expected time of when the request was created might not always be 100% accurate. But making it 100% accurate would require a lot of effort I think.

So basically I would propose that you create a function similar to calculateDeadline(), that instead of having a `period`, it takes the current `activeBlock` number and the `opentime`, and then it subtracts them to get the difference which you can then use to derive the expected date that it happened with the block time of 12 seconds (assuming the happy-path that the active block counter was incremented every block). 

Does it make sense?

gonzamontiel:
> activeBlock



> @gonzamontiel good point. You would have to implement something similar to the [calculateDeadline](https://github.com/pendulum-chain/portal/blob/d92c3abe36dac9e40b416ea8d51b5051618f1ebd/src/helpers/spacewalk.ts#L76)() function.
> 
> The `opentime` is indeed not the block hash but it's the active block number that was the latest when the request was issued. Note that, the active block number is an extra counter that spacewalk has, which is only incremented if the oracle is running. If the oracle is offline because price data is not up-to-date, then this active block counter is not incremented per block. This is some detail that I would omit for simplicity, however. Thus, what you show as the expected time of when the request was created might not always be 100% accurate. But making it 100% accurate would require a lot of effort I think.
> 
> So basically I would propose that you create a function similar to calculateDeadline(), that instead of having a `period`, it takes the current `activeBlock` number and the `opentime`, and then it subtracts them to get the difference which you can then use to derive the expected date that it happened with the block time of 12 seconds (assuming the happy-path that the active block counter was incremented every block).
> 
> Does it make sense?

Thanks for the detailed explanation and suggestion. If I unfderstood correctly, what we have is already something like "blocks ago". To convert that to time ago, we would only need to multiply that by 12. Then, substract that number of seconds to DateTime.now(). Does that make sense? It's what I implemented.

src/helpers/spacewalk.ts
```diff
@@ -87,3 +87,12 @@ export function calculateDeadline(
...
   return end;
 }
+
+export function estimateRequestCreationTime(
+  activeBlockOpenTime: number,
+  blockTimeSec = 12,
+) {
+  const secondsAgo = activeBlockOpenTime * blockTimeSec;
+  const now = DateTime.now();
+  return now.minus({ seconds: secondsAgo })
```
ebma, line 97: This would not really work as expected. The problem is that `activeBlockOpenTime` is not related to the current 'epoch' at the time but it's just some number that is incremented starting from 0 when the blockchain started. Thus, we can't relate it to `DateTime` that easily. 

I think the following should work
```
export function estimateRequestCreationTime(
  activeBlockOpenTime: number,
  currentActiveBlock: number,
  blockTimeSec = 12,
) {
  const activeBlocksPassed = currentActiveBlock - activeBlockOpenTime;
  const secondsAgo = activeBlocksPassed * blockTimeSec;
  const now = DateTime.now();
  return now.minus({ seconds: secondsAgo })
}
```

gonzamontiel: Uhm I see, but how often is the `activeBlockOpenTime` incremented? If it's a counter that gets incremented per block? If that is the case, then it's exactly the same as saying that the it started N blocks ago. From your code I would understand that `activeBlockOpenTime` is exactly the block number when the request was created. Then, subtracting the current block number minus the block number when the request started, you indeed get an amount of blocks. So `activeBlockOpenTime` is a `BlockNumber`. Is that correct ?

gonzamontiel: Okay, that is correct according to [this](https://github.com/interlay/interbtc/blob/b174d63da831cadff70c117633c8a0cb4af790cc/primitives/src/lib.rs#L129). I am still getting some weird dates using your code. I am getting the current active block from getActiveBlockNumber() in useSecurityPallet()

ebma: > So activeBlockOpenTime is a BlockNumber. Is that correct? 

yes exactly

What are the values you pass for `activeBlockOpenTime` and `currentActiveBlock`, did you log them and can maybe share them to get a better understanding?

ebma (commented):
Looks good to me so far 👍 

But one question, are you also going to include the dialogs that are requested in the issue ticket or do you want to do that in another branch? I checked and the ticket says

> On clicking any of the list item.
Show a new page with additional details of the transaction.

And the Figma project does contain some designs for these dialogs.

src/pages/bridge/TransfersColumns.tsx
```diff
@@ -0,0 +1,56 @@
+import { ColumnDef } from '@tanstack/table-core';
+
+export type TransferStatus = 'Pending' | 'Completed' | 'Cancelled' | 'Reimbursed' | 'Retried';
+
+export enum TransferType {
+  issue = 'issue',
+  redeem = 'redeem',
+}
+
+export interface TTransfer {
```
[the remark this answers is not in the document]
gonzamontiel: It stands for Type, I only used the same pattern as @nejcm used for Collators. Because the table is parametrizable <T>

gonzamontiel:
> Looks good to me so far +1
> 
> But one question, are you also going to include the dialogs that are requested in the issue ticket or do you want to do that in another branch? I checked and the ticket says
> 
> > On clicking any of the list item.
> > Show a new page with additional details of the transaction.
> 
> And the Figma project does contain some designs for these dialogs.
> 
> Ah, and can we maybe make the table sort the items descending by the 'updated' column by default?

Oh, you are actually right! the details dialog is missing. I will also include the order by udpated at loading.

src/pages/bridge/TransfersColumns.tsx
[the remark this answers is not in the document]
nejcm: Yeah usually a convention I use. Most of the time for generics.

gonzamontiel:
hey @prayagd I saw you updated the figma files, please do let me know when I can check them again.

gonzamontiel:
hey @prayagd , also, for all statuses, I am including this same gray box: 
Screenshot from 2023-04-05 17 14 58

I took it as the "common info" among all transfers, so I propose that we agree in the data to show there no mater the status, so far, what I show is in the image above.

gonzamontiel:
I added an important note to the description, so you are aware @ebma @prayagd

ebma (commented):
I pushed some changes. For some things I was not yet sure how exactly the implementation should look so I went ahead and added code instead of first commenting. 

I also changed the logic back to show 'cancelled' instead of 'pending' for timed out requests. I had a look how it feels and have to say that telling the user that it's cancelled is probably better because otherwise he might think that there is still an action to be taken (and the dialog also has some issues because we then count up with negative numbers instead of counting down as it's already in the past). Sorry for the back and forth here @gonzamontiel. 

Some remaining remarks:
- If we don't show dialogs for 'retried' requests let's not show a button for them.
- I would change the button icon for opening the details dialog. I think a lense or some 'document' icon would make more sense. The current one suggests that you are about to open a new tab/open an external link. Actually, it might be best if you open the dialog without the extra button but just by clicking on the whole table row. Not sure how this was intended originally, maybe you can clarify @prayagd 
- The 'reimbursed' dialog should indicate an amount that the user got extra (the punishment fee). But we can only estimate it with the current exchange rates which will result in loss of precision and showing wrong values. @prayagd does interlay show a dialog for reimbursed requests? If so, what amounts do they show?
- We should remove the on-click link for 'Foucoco opentime block' as it will never link to the correct block. We just can't link properly without an indexer

Also, maybe we should swap the amount precision in table vs dialog. ATM ,the table will show more precise amounts than the details-dialog which is odd because as a user I would expect the dialog to show me the most accurate information.
image

gonzamontiel:
> I pushed some changes. For some things I was not yet sure how exactly the implementation should look so I went ahead and added code instead of first commenting.
> 
Thanks!

> I also changed the logic back to show 'cancelled' instead of 'pending' for timed out requests. I had a look how it feels and have to say that telling the user that it's cancelled is probably better because otherwise he might think that there is still an action to be taken (and the dialog also has some issues because we then count up with negative numbers instead of counting down as it's already in the past). Sorry for the back and forth here @gonzamontiel.

Hey I'm sorry about the confusion. I saw you uncommented the logic that was already there, but I mistakenly left commented ( and added a harcoded 'Pending', but this was only to force some pending transactions, bc there weren't any). But the whole idea was to show Cancelled, I implemented it that way and we were on the same page :ok_hand: 

> Some remaining remarks:
> 
> * If we don't show dialogs for 'retried' requests let's not show a button for them.

Yup, there are no designs, so I guess I can hide the button.

> * I would change the button icon for opening the details dialog. I think a lense or some 'document' icon would make more sense. The current one suggests that you are about to open a new tab/open an external link. Actually, it might be best if you open the dialog without the extra button but just by clicking on the whole table row. Not sure how this was intended originally, maybe you can clarify @prayagd
Yeah, I was lazy on that one and included the first that came to my mind, but regarding the whole row thing, I wrote in the description that "I included an extra column to see the details (involving the whole row was inconvenient, and I think this solution works pretty well)". This is because we are reusing the Table component, and when I tried the refactor (adding an anchor for the row) the whole table looked very ugly.  I think this can be an improvement for the future, so far, let's change the icon :ok_hand: 

> * The 'reimbursed' dialog should indicate an amount that the user got extra (the punishment fee). But we can only estimate it with the current exchange rates which will result in loss of precision and showing wrong values. @prayagd does interlay show a dialog for reimbursed requests? If so, what amounts do they show?

I am not sure about this one so I leave ti to @prayagd 

> * We should remove the on-click link for 'Foucoco opentime block' as it will never link to the correct block. We just can't link properly without an indexer

Okay, so it actually doesn't mean anything to the chain itself, it is not a block number finally.

> Also, maybe we should swap the amount precision in table vs dialog. ATM ,the table will show more precise amounts than the details-dialog which is odd because as a user I would expect the dialog to show me the most accurate information. image

OH, that is on the format() function to blame. I wrote that one to simplify showing extra large amount of numbers in a shorter way, but idk if it is actually required, I could show the whole number too.

ebma:
Just a minor thing I noticed is that the dialogs don't have a smooth fade-in/out transition. If you compare that to the dialogs we have for the collator list, for example, they have some transitions. I would guess that it's because they are only mounted when they are needed instead of being mounted 'all the time' and toggled by the `visibility` prop. As I said just a minor thing but nice to have. @prayagd can you address the questions that @gonzamontiel and I raised in the last comments?

gonzamontiel:
> Just a minor thing I noticed is that the dialogs don't have a smooth fade-in/out transition. If you compare that to the dialogs we have for the collator list, for example, they have some transitions. I would guess that it's because they are only mounted when they are needed instead of being mounted 'all the time' and toggled by the `visibility` prop. As I said just a minor thing but nice to have. @prayagd can you address the questions that @gonzamontiel and I raised in the last comments?

So, the implementation was on purpose, buuut I did not notice that the transitions were missing and that it owuld be an issue (?). I guess I could change it but I'd rather prefer to merge this for good :)

ebma:
Would be okay to merge from my side but I'd want @prayagd to acknowledge the minor issues and have the last word for merging.

prayagd:
@ebma @gonzamontiel First of all thank you for the detailed review.
Regarding the earlier comments, the documents icon looks good. 
Regarding the reimburse dialog, yes interlay shows the dialog for reimburse along with KSM/DOT received by burning the bridged asset and also the compensation fee.

ebma:
I found the code that interlay uses for that [here](https://github.com/interlay/interbtc-ui/blob/47fcc02687f9e31c1fe842126d0b862ac78ba10d/src/legacy-components/RedeemUI/RedeemRequestStatusUI/ReimbursedRedeemRequest/index.tsx#L68-L72). They indeed convert the amount using the current price. 

We can also convert the prices between the assets with the on-chain price info by calling the oracle pallets RPC methods. The oracle pallet [exposes](https://github.com/pendulum-chain/spacewalk/blob/main/pallets/oracle/rpc/src/lib.rs) the `currency_to_usd()` and `usd_to_currency()` functions, which can be used to convert from eg KSM to USDC by first calling `currency_to_usd(amount, KSM)` for KSM and then use the result with `usd_to_currency(amount, USDC)` to get the respective value in the other currency. (Note that I simplified the function calls. 

@prayagd can you please clarify the things you want to have before we merge this? Is showing the better version of the reimbursement important? Is showing working buttons for cancelling the redeem request (reimbursement yes/no) important?

gonzamontiel:
> I found the code that interlay uses for that [here](https://github.com/interlay/interbtc-ui/blob/47fcc02687f9e31c1fe842126d0b862ac78ba10d/src/legacy-components/RedeemUI/RedeemRequestStatusUI/ReimbursedRedeemRequest/index.tsx#L68-L72). They indeed convert the amount using the current price.
> 
> We can also convert the prices between the assets with the on-chain price info by calling the oracle pallets RPC methods. The oracle pallet [exposes](https://github.com/pendulum-chain/spacewalk/blob/main/pallets/oracle/rpc/src/lib.rs) the `currency_to_usd()` and `usd_to_currency()` functions, which can be used to convert from eg KSM to USDC by first calling `currency_to_usd(amount, KSM)` for KSM and then use the result with `usd_to_currency(amount, USDC)` to get the respective value in the other currency. (Note that I simplified the function calls.
> 
> @prayagd can you please clarify the things you want to have before we merge this? Is showing the better version of the reimbursement important? Is showing working buttons for cancelling the redeem request (reimbursement yes/no) important?

I think is not a matter of importance, I'm pretty sure _it is important_, but we need to think in a different variable which is the time this PR was open. We need to move forward. The feature is not even visible yet. We can iterate on this. Even more, the ticket initially didin't even spoke about any of this. So let's be fair, let's collect any pending matters of this issue, and create a new ticket to address them (@prayagd ). Then, let's merge this _as is_ :pray: . Thanks

prayagd:
@ebma I assume the expired redeem request has both buttons for reimbursement called as "Reimburse" or letting the user try the redeem request again by clicking "Compensate". Correct @gonzamontiel ?

@ebma if we are not showing the on-chain price then what values are we showing? Only the amount of the asset? It would be very helpful if you could attach a screenshot

Agree with @gonzamontiel Lets go ahead with the PR and merge, anyway this is going to be a community launch for Spacewalk. We can always iterate on the things we dont find are correct and as wrt to on-chain prices As i understand this is about only punishment fee which would be in KSM/DOT. Lets go ahead to only show the amounts until we figure out a way to precisely calculate the values. wdyt @ebma ?

ebma:
@prayagd you can see what it looks like in the deploy preview. We have a reimbursed request on Foucoco, so you should be able to find it in the table [here](https://deploy-preview-91--rococo-souffle-a625f5.netlify.app/foucoco/transfers).

It's fine for me to merge it and add more tweaks afterward. We just need to be aware of what is left to change and create a proper follow-up ticket.

So I guess we agreed to merge this as is now. We need to resolve the conflicts first though @gonzamontiel

prayagd:
@ebma ah thanks, i was checking it on Amplitude. Hence, could not find it.

prayagd:
@ebma @gonzamontiel Tickets for the next iteration
1. https://github.com/pendulum-chain/portal/issues/103
2. https://github.com/pendulum-chain/portal/issues/104
Will add more as we go along.

gonzamontiel:
hey @ebma , it is fixed, let's approve and merge this one?

ebma (approved):
LGTM and ready for merge

PR #93, opened 2023-03-23
92 change stellar amount handling to accommodate for 12 decimals

### LINKED ISSUE #92: Change Stellar amount handling to accomodate for 12 decimals
We recently changed the decimals for Stellar amounts from 7 to 12 decimals on Spacewalk. The values are up-scaled now, but this has to be taken into account when calling the respective extrinsics and when converting between units in the bridge UI.

ebma:
This changes the value for StellarDecimals from 7 to 12. This is necessary because we recently changed this in the spacewalk pallets. 
Furthermore, we modified the memo that needs to be added to a Stellar transaction in order to fulfill an issue request. The dialog now shows the reduced version of the memo. 

Closes #92.


gonzamontiel (approved):

ebma:
@gonzamontiel can you re-review my latest commits? I changed it so that the wsprovider does not auto connect to amplitude. I figure that the auto connect causes more harm than good because it introduced this nasty race condition when connecting to another endpoint before the other connection is fully established. It also does not really make sense that it auto connects to amplitude regardless of the current url. 
The reason why I changed it here is because it introduced weird behaviour for the vault selection because if I am testing on the local chain and it is not running, then instead of not showing any vaults as would be expected, it showed the amplitude vaults instead.

gonzamontiel (approved):

gonzamontiel:
This is ready for merging @ebma

PR #95, opened 2023-03-30
Add total issuance stats & others

### LINKED ISSUE #89: Output token stats on portal
### User story / Requirement
As a consumer of token analytics (like CoinGecko), I'd like to access the current state of the $PEN and $AMPE token.

### Solution option
1. There is a [script](https://github.com/pendulum-chain/bi-scripts) that extracts the following token parameters for the PEN token:

- the total amount of tokens
- the total amount of transferable tokens (tokens in circulation)
- the total locked amount of tokens
- the total reserved amount of tokens

2. Output the result of the script on a new page on website / portal. See example [here](https://apps.moonbeam.network/moonbeam/stats/).

Question: Can this output be in real time i.e. updated for each block or should I suggest a frequency of update e.g. once every 1000 blocks?

gonzamontiel:
- Results are shown in `/<chain>/stats`
- I _had to_ include it inside the Portal since we already have a Router in place (it was the easiest)
- I added the mininum design to it
Closes #89

### DIFF
--- a/src/app.tsx
+++ b/src/app.tsx
@@ -8,6 +8,7 @@ import { Amm } from './pages/amm/Amm';
 import Bridge from './pages/bridge';
 import { Collators } from './pages/collators/Collators';
 import { Dashboard } from './pages/dashboard/Dashboard';
+import { Stats } from './pages/stats';
 
 export function App() {
   return (
@@ -19,6 +20,7 @@ export function App() {
           <Route path="amm" element={<Amm />} />
           <Route path="bridge/:to?" element={<Bridge />} />
           <Route path="collators" element={<Collators />} />
+          <Route path="stats" element={<Stats />} />
           <Route path="*" element={<NotFound />} />
         </Route>
       </Routes>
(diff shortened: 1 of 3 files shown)


ebma:
Looks good. But who actually is gonna be consuming this information @vadaynujra? Is it going to be an application that automatically queries these stats from our page? Or is it gonna be humans trying to get an overview? Because if it's to be used by machines it would make more sense to approach this differently and just try to return a JSON response for a GET request.

gonzamontiel:
> Looks good. But who actually is gonna be consuming this information @vadaynujra? Is it going to be an application that automatically queries these stats from our page? Or is it gonna be humans trying to get an overview? Because if it's to be used by machines it would make more sense to approach this differently and just try to return a JSON response for a GET request.

I think this closes the ticket as is described. It doesn't say it should be a feed.
A different requirement can be handled in another ticket. We can benefit from this one because calculations are already done. So I would approve the PR _as is_

ebma:
> I think this closes the ticket as is described. It doesn't say it should be a feed.
True. LGTM then

ebma (approved):

PR #98, opened 2023-04-06
fix: wallet persistence

nejcm:
Take a look if this fixes the wallet issue after reload. @gonzamontiel @ebma


src/GlobalStateProvider.tsx
```diff
@@ -70,16 +62,46 @@ const GlobalStateProvider = ({
...
+  });
+
+  useEffect(() => {
+    const run = async () => {
+      if (!account) return;
+      const name = storageService.get('@talisman-connect/selected-wallet-name');
+      if (!name) return;
+      const wallet = getWalletBySource(name);
+      if (!wallet) return;
+      // TODO: optimize this
```
ebma, line 80: How? Or what exactly do you want to optimize?

ebma (commented):
Can confirm this indeed fixes the problem, thanks a lot 🎉 

Something unrelated but annoying are the changes in the yarn.lock file. I think I am part of the problem because my yarn will always create a lock file for version 2 and yours then creates another one instead. Do you have some suggestion how we can improve this in the future so that we don't have to change it again and again @nejcm? Should we just agree on a version?

src/GlobalStateProvider.tsx
nejcm: Problem here is that is beinging used in multiple parts of the app and just want to create something reusable and standardized.

nejcm: Regarding yarn we can in package json define what version should be used and that will be enforced than. What version are you guys using and lets agree on that.

ebma: Is there a good reason to stick to yarn v1 lock files? 
My `yarn --version` says `4.0.0-rc.38`

nejcm: Not really. I am working eith some projects that use v1 as its still more wide spread. Bilut yeah we can upgrade here

ebma:
Can you then regenerate the yarn.lock file for version 2 so that there are not so many changes @nejcm? 

Also it seems that there are some lint complaints we need to fix.

nejcm:
Done. I reverted the yarn.lock file. Will leave it up to you guys to configure the version

gonzamontiel (approved):
I fix an unrelated little issue that prevented me to test properly. I can also confirm this  fixes the issue :pray:

PR #102, opened 2023-04-20
chore: fix release action

nejcm:
I am trying to fix that failing action that will automatically increment the app package version and publish a changelog.
Now for that because the branch is protected a separate github token must be created with proper perms by somebody that has admin right on the repo/organization and that add it to the repo secrets.

Call it GH_TOKEN with perms listed in this link:
https://github.com/semantic-release/github#github-authentication

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,4 +1,4 @@
-name: push
+name: release
 on:
   push:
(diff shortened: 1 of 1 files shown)


gonzamontiel (approved):
thanks for the fix!

PR #109, opened 2023-04-25
Fetch collator identities

gonzamontiel:
- Added usePalletIdentity hook
- Added a function to fetch the identity info of a single account address
- Pre fetch identities for all collators to show in the table
  - **Note**: this slows down the loading of the table a bit, if we want to make it non-blocking, then the collators name column would remain empty for some time which is also bad UX.
-  Collator column shows (in the following order, if first is missing): display > email > twitter > account id

### DIFF
--- a/src/hooks/useIdentityPallet.ts
+++ b/src/hooks/useIdentityPallet.ts
@@ -1,0 +1,40 @@
+import { useMemo } from 'preact/hooks';
+import { useNodeInfoState } from '../NodeInfoProvider';
+
+export interface PalletIdentityInfo {
+  email?: string;
+  riot?: string;
+  twitter?: string;
+  web?: string;
+  display?: string;
+}
+
+export function useIdentityPallet() {
+  const { api } = useNodeInfoState().state;
+
+  const memo = useMemo(() => {
+    return {
+      async identityOf(account: string): Promise<PalletIdentityInfo | undefined> {
+        if (!api || !api.query.identity) {
+          return;
+        }
+        const indentityResponse = await api.query.identity.identityOf(account);
+        if (indentityResponse.isEmpty) {
+          return;
+        }
+        // Casting the data to the real value PalletIdentityRegistration, fails to show readable info
+        // Therefore the cast to 'any'
+        const pir: any = indentityResponse.toHuman();
+        return {
+          email: pir.info.email ? pir.info.email.Raw : '',
+          riot: pir.info.riot? pir.info.riot.Raw : '',
+          twitter: pir.info.twitter? pir.info.twitter.Raw: '',
+          web: pir.info.web? pir.info.web.Raw : '',
+          display: pir.info.display? pir.info.display.Raw : '',
+        };
+      }
+    };
+  }, [api]);
+
+  return memo;
+}
--- a/src/hooks/useIdentityPallet.ts
+++ b/src/hooks/useIdentityPallet.ts
@@ -1,12 +1,14 @@
+import { PalletIdentityRegistration } from '@polkadot/types/lookup';
+import { hexToU8a } from '@polkadot/util';
 import { useMemo } from 'preact/hooks';
 import { useNodeInfoState } from '../NodeInfoProvider';
 
 export interface PalletIdentityInfo {
+  display?: string;
   email?: string;
   riot?: string;
   twitter?: string;
   web?: string;
-  display?: string;
 }
 
 export function useIdentityPallet() {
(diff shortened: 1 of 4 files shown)


ebma (commented):
The code looks good. 

But I think it's problematic that as a power-user, I am now not able to see the actual address of each collator anymore. Seeing the identity is great but maybe we could show the address in 'secondary'-font below the fetched identity? Or if that's too ugly, maybe at least make the identity name a clickable text, that stores the address of that collator to the clipboard on-click.

prayagd:
Agree to @ebma suggestion. It would be very good ux, if we could copy the address by clicking on the name of the collator. @gonzamontiel Let us know if you can implement it in this PR itself. I know this is not specified in this ticket so also happy to create a new ticket for this. We can keep shipping tiny bits.

gonzamontiel:
> Agree to @ebma suggestion. It would be very good ux, if we could copy the address by clicking on the name of the collator. @gonzamontiel Let us know if you can implement it in this PR itself. I know this is not specified in this ticket so also happy to create a new ticket for this. We can keep shipping tiny bits.

Uhm. I can show a modal with extra info of the collator, like email, twitter, etc, if present. I mean it is also bad ux that you click on a name and something gets copied, not intuitive. better to open a modal, then there, we can have a clickable address.

prayagd:
@gonzamontiel How about a modal but just with address and nothing more. I feel showing all other information wont be relevant there and also i dont think so users are interested in knowing the twitter address of a collator. If this requirement comes from user end we can then append all these details in the modal too. what say?

ebma:
But what's wrong with just showing the address below the identity name? Don't you think creating a modal dialog just for showing the 'real' address is quite an overkill? We can also re-use the clickable address component so that it's copyable.

gonzamontiel:
> > Don't you think creating a modal dialog just for showing the 'real' address is quite an overkill? We can also re-use the clickable address component so that it's copyable.
> 
> If only showing an address, then I'd go for your option @ebma, showing it below and smaller, reusing the clickable address. For me what was not acceptable is to click on the name and copy an address. Will implement this and let you know, thanks for the input.

Only downside, for complex cells like this one we lose the orderBy functionality.
Alternatively, we can have a different column for the clickable address. wdyt @ebma @prayagd ?

src/hooks/useIdentityPallet.ts
```diff
@@ -0,0 +1,43 @@
...
+export function useIdentityPallet() {
+  const { api } = useNodeInfoState().state;
+
+  const memo = useMemo(() => {
+    return {
+      async identityOf(account: string): Promise<PalletIdentityInfo | undefined> {
+        if (!api || !api.query.identity) {
+          return;
+        }
+        const indentityResponse = await api.query.identity.identityOf(account);
```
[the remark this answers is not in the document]
gonzamontiel: fixed

prayagd:
@gonzamontiel IMHO we dont need a new column maybe the address just next to the name with the copy icon. So users can click on the icon and copy and if the collators does not have any name then we can go with "Unknown identity".

gonzamontiel:
> @gonzamontiel IMHO we dont need a new column maybe the address just next to the name with the copy icon. So users can click on the icon and copy and if the collators does not have any name then we can go with "Unknown identity".

Uhm okay, so we lose sorting (we cannot sort by collator name, without some refactor on the Table logic).

Option 1
Screenshot from 2023-04-27 15 14 41

Option 2
Screenshot from 2023-04-27 15 16 24

prayagd:
@gonzamontiel my vote goes for option 2

gonzamontiel:
> @gonzamontiel my vote goes for option 2

done.

gonzamontiel:
> Looks good to me but maybe we should change the orderBy behavior as you pointed out @gonzamontiel. We can do that by simply changing the accessor key to the identity name if I'm not mistaken?
> 
> At least I would now expect that if I click on the column to sort by names, it sorts by the identity name and not by the address. This is just a nit though.

Maybe I was not clear enough when I said we would be _losing_ it. I added the current implementation of sortBy to the Table, that works with simple columns defined by accessorFn, which returns a _string_.
If we replace the whole cell, returning a `<div>` or anything different than text, then the sorting function doesn't know how to sort that. I checked and `react-table` offers multiple ways of customizing sorting (like providing a custom a sortingFn), but they are not implemented yet.

gonzamontiel:
> I'm still not sure I understand the problem. AFAIK react-table would use the `accessor` of a column for sorting, and if you want to customize that you'd only need to change the `accessorFn()`, like is done [here](https://github.com/pendulum-chain/portal/blob/96-fetch-collator-identities/src/components/Pools/Swap/columns.tsx) (?) Since you can access the whole data object in that `accessorFn` you can transform it as you wish and derive the collator name, which would then be used for sorting I suppose. The jsx we put into the `cell` should not matter for sorting.

Somehow I thought the `accesorFn` and the `cell` properties were exclusive (!). I mean, `accesorFn` without `cell`, will output to the UI, while if you specify the cell property, the `accessorFn` has _apparently_ no impact. But I just found that from what you say that allows ordering. My bad, it is fixed now.

src/pages/collators/columns.tsx
```diff
@@ -28,6 +28,7 @@ const getAmountDelegated = (candidate: ParachainStakingCandidate, address: strin
 export const nameColumn: ColumnDef<TCollator> = {
   header: 'Collator',
   accessorKey: 'collator',
+  accessorFn: ({ identityInfo }) => identityInfo?.display || '0',
```
ebma, line 31: Maybe we should replace the `0` with the actual text that you display in case it is empty? This is just nitpicking but it's a bit weird that the 'Unknown' collators seem to not follow the alphabetic ordering 

```suggestion
  accessorFn: ({ identityInfo }) => identityInfo?.display || 'Unknown',
```

ebma (approved):
LGTM, just pointed out something minor

src/pages/collators/columns.tsx
gonzamontiel: oh, maybe this is not too intuitive. I prefer to have all the meaningful names together when ordering (from A to Z) and then Unkwnon as a special category. Since it is an arbitrary word, its order doesn't matter.

PR #120, opened 2023-04-26
75 accept tcs pop up

gonzamontiel:
- Added an 'expire' field to useLocalStorage() hook
- Added terms and conditions dialog (looks ok in both themes)
- Slightly changed the texts.


src/TermsAndConditions.tsx
```diff
@@ -0,0 +1,43 @@
...
+    set('accepted');
+  };
+
+  return state ? (
+    <></>
+  ) : (
+    <Modal open={true} style={{ borderRadius: '5px' }}>
+      <Modal.Header className="text-3xl">T&C</Modal.Header>
+      <Modal.Body>
+        <div className="text-lg mb-4">Alternative links can be found in the docs.</div>
```
ebma, line 22: This text is weird. Was that like that in some design?

ebma (changes requested):
I can't click on 'Agree' if the checkbox for not showing the dialog again is not checked. This doesn't make sense as I should be able to agree regardless.

prayagd:
@ebma Agree. Maybe add a underline to the "Terms and conditions" text?
Also @gonzamontiel can you please remove this text "Alternative links can be found in the docs"? Looks like it was not updated by Marina.

src/TermsAndConditions.tsx
gonzamontiel: Is super weird, it was part of the designs.

gonzamontiel:
> @ebma Agree. Maybe add a underline to the "Terms and conditions" text? Also @gonzamontiel can you please remove this text "Alternative links can be found in the docs"? Looks like it was not updated by Marina.

@prayagd , let's double check the texts please :pray: , and to reinforce, in this case I would include them in the description of the ticket.

gonzamontiel:
> > I can't click on 'Agree' if the checkbox for not showing the dialog again is not checked. This doesn't make sense as I should be able to agree regardless.
> 
> Again, you're right I guess I blindly follow the designs on this one, will remove the condition.
> 
> Screenshot from 2023-04-27 11 30 08

But thinking again about it, there is something weird about the design. The user woudl always chose this option. If not, then nothing is saved in storage, and the Agree button only acts as a Close button (nothing changes in the state, and the dialog is closed)
So, even we leave this apparently weird condition, or we remove the checkbox and we notify the user it will be stored for 30 days. @prayagd please decide here.

prayagd:
@gonzamontiel You are right here, let me get back to you here.

prayagd:
@gonzamontiel After carefully looking into it i think it would need a new design and change of text, If i may suggest should we put this on hold and will get you new designs and change of text wdyt?

erasmus:
What is the purpose of the "Remember for 30 days"? What's actually important is which version of the T&C they have accepted... Much better would be to store the date/version of acceptance

prayagd:
@erasmus The purpose of 30 days was not make the user accept T&Cs every-time he connects to the portal and store just the acceptance to t&cs. Will change the main ticket for version storing and get the designs also updated.

src/hooks/useLocalStorage.ts
```diff
@@ -60,6 +63,9 @@ export const useLocalStorage = <T>({
     (value) => {
       storageSet(key, value);
       setState(value);
+      if (expire) {
+        debounce(storageService.remove, expire * 60)(key);
+      }
```
ebma, line 68: I'm not sure if I understand this logic. If we want to make this value expire after some period of time, then we'd have to store the expiry in an additional timestamp item in local storage instead of using it like that. This would now only work if you keep the app open. 
Also, why multiply with `60`? The `expire` value is in seconds but the `debounce` value is in milliseconds so we would have to multiply with `1000`.

ebma (changes requested):

src/hooks/useLocalStorage.ts
gonzamontiel: You are totally right, this only works with the app working. I'm not sure how the 60 got there, probably I was testing and needed a lower value to see the effects, but weird.
I removed this functionality from this branch, but created another PR with an improved version here https://github.com/pendulum-chain/portal/pull/186

gonzamontiel: BTW, we don't need the expire date anymore, so it wasn't needed anyways here.

ebma: Got it, thanks :+1:

ebma (approved):

PR #125, opened 2023-05-02
77 dropdown to switch between versions of portal

### LINKED ISSUE #77: Dropdown to switch between versions of Portal
As a portal user, I should be able to switch between portal version's using a dropdown.

# Acceptance criteria

* Show network dropdown next to wallet connect
* Show three options in this dropdown
    * Amplitude
    * Pendulum
    * Foucoco
* Selecting any option should direct the user to respective URL of the portal
*  Default behaviour of the portal should be https://portal.pendulumchain.org/pendulum/ 

# Wireframe
https://www.figma.com/file/5up5sghcttBogehUq4255i/App?node-id=1782-1232&t=wW8viFRDexb8HqRW-0

gonzamontiel:
Closes #77

- I updated the logos based on Figma
- Selecting a different chain redirects to the dashboard of the chain (no matter where you are)
-  Pen icon on black background looks with almost no contrast, but there is no design for that I guess, and we can live with it for now


src/components/Layout/index.tsx
```diff
@@ -11,6 +11,7 @@ import NetworkId from './NetworkId';
 import SocialAndTermLinks from './SocialAndTermLinks';
 import Versions from './Versions';
 import './styles.sass';
+import ChainSeletor from '../ChainSelector';
```
ebma, line 14: Nit: Typo

src/components/ChainSelector.tsx
```diff
@@ -0,0 +1,51 @@
...
+      </Dropdown.Menu>
+    </Dropdown>
+  </>
+);
+
+function toTitle(str: string) {
+  return str.charAt(0).toUpperCase() + str.slice(1);
+}
+
+export default ChainSeletor;
```
ebma, line 51: Nit: typo

ebma (commented):
The selector looks good. Few things that I noticed:
- I would reduce the gap between the tenant name and the arrow down icon 
- The Pendulum logo seems bigger than the Amplitude logo. Maybe it's just because the Amplitude logo has some padding in itself but it's weird
- I think always navigating to the dashboard is not good UX. If I as a user am currently on the bridge panel and decide to switch the networks, I probably want to stay on the bridge panel. Otherwise, I would navigate back to the dashboard by myself anyways. @prayagd what do you think about this?

prayagd:
@gonzamontiel wrt to navigation when changing network it would be good if the users stays on the same feature even if the network is changed and not go back to dashboard.

gonzamontiel:
> @gonzamontiel wrt to navigation when changing network it would be good if the users stays on the same feature even if the network is changed and not go back to dashboard.

gotcha, will change it asap :)

gonzamontiel:
> The selector looks good. Few things that I noticed:
> 
> * I would reduce the gap between the tenant name and the arrow down icon

Well I followed the designs (and made the space smaller than it actually is) I guess is just a matter of taste. Anyway I wasn't 100% accurate with the designs bc the dropdwown menu looks pretty okay to my taste without the need of customizing a lot.

> * The Pendulum logo seems bigger than the Amplitude logo. Maybe it's just because the Amplitude logo has some padding in itself but it's weird
I can see something there, will see what I can do.
=

gonzamontiel:
Following your comments I dedicated a bit more to make it nicer and closer to the designs, here how it looks:
Screenshot from 2023-05-03 16 48 55
Screenshot from 2023-05-03 16 48 10

ebma (approved):
LGTM otherwise

prayagd (approved):
Good to go

src/components/Layout/index.tsx
gonzamontiel: jeeez, you can call it whatever you like when is the default haha. fixed

PR #139, opened 2023-05-09
Fix/136 Add support for XLM vaults

gonzamontiel:
- Removed warning about issuer for XLM
-  Removed 'issued by' for XLM requests

### DIFF
--- a/src/helpers/spacewalk.ts
+++ b/src/helpers/spacewalk.ts
@@ -131,7 +131,7 @@ export function currencyToString(currency: SpacewalkPrimitivesCurrencyId, tenant
 function tryConvertCodeToAscii(code: U8aFixed) {
   const ascii = hex_to_ascii(code.toHex());
   if (ascii !== ascii.trim()) {
-    throw Error('Asset has invalid characters');
+    throw Error('Asset code contains invalid space characters');
   }
   return ascii.replace('\0', '');
 }
--- a/src/helpers/spacewalk.ts
+++ b/src/helpers/spacewalk.ts
@@ -1,11 +1,12 @@
 import bs58 from 'bs58';
 import { H256 } from '@polkadot/types/interfaces';
 import { ApiPromise } from '@polkadot/api';
-import { SpacewalkPrimitivesCurrencyId } from '@polkadot/types/lookup';
+import { SpacewalkPrimitivesAsset, SpacewalkPrimitivesCurrencyId } from '@polkadot/types/lookup';
 import { DateTime } from 'luxon';
 import { Asset, Keypair } from 'stellar-sdk';
 import { convertRawHexKeyToPublicKey } from './stellar';
 import { TenantName } from '../models/Tenant';
+import { U8aFixed } from '@polkadot/types-codec';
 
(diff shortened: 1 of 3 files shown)


src/helpers/spacewalk.ts
```diff
@@ -98,11 +97,11 @@ export function currencyToString(currency: SpacewalkPrimitivesCurrencyId, tenant
...
-      const code = hex_to_ascii(stellarAsset.asAlphaNum4.code.toHex());
+      const code = hex_to_ascii(stellarAsset.asAlphaNum4.code.toHex()).trim();
       const issuer = convertRawHexKeyToPublicKey(stellarAsset.asAlphaNum4.issuer.toHex());
       return `${code}:${issuer.publicKey()}`;
     } else if (stellarAsset.isAlphaNum12) {
-      const code = hex_to_ascii(stellarAsset.asAlphaNum12.code.toHex());
+      const code = hex_to_ascii(stellarAsset.asAlphaNum12.code.toHex()).trim();
       const issuer = convertRawHexKeyToPublicKey(stellarAsset.asAlphaNum12.issuer.toHex());
       return `${code}:${issuer.publicKey()}`;
     } else {
```
ebma, line 107: Let's maybe wrap the content in the currencyToString functin with a try-catch as well.

src/helpers/spacewalk.ts
```diff
@@ -35,12 +35,11 @@ export function convertCurrencyToStellarAsset(currency: SpacewalkPrimitivesCurre
...
       return Asset.native();
     } else if (stellarAsset.isAlphaNum4) {
-      const code = hex_to_ascii(stellarAsset.asAlphaNum4.code.toHex());
-      console.log('code', code);
+      const code = hex_to_ascii(stellarAsset.asAlphaNum4.code.toHex()).trim();
       const issuer = convertRawHexKeyToPublicKey(stellarAsset.asAlphaNum4.issuer.toHex());
       return new Asset(code, issuer.publicKey());
     } else if (stellarAsset.isAlphaNum12) {
-      const code = hex_to_ascii(stellarAsset.asAlphaNum12.code.toHex());
+      const code = hex_to_ascii(stellarAsset.asAlphaNum12.code.toHex()).trim();
```
ebma, line 42: `trim()`ing is problematic because we then support exactly those assets that are 'faulty'. 

We will soon have the assets `BRL `, `TZS ` and `BRL\0`, `TZS\0` on-chain. 
What we want to do is, ignore `BRL `, `TZS ` (ie. the assets with a space character) and support `BRL\0`, `TZS\0`. Thus, instead of using `trim()` we should instead `replace('\0','')` if necessary. We should also reject all asset codes that contain space characters, just to be on the safe side. In these cases we can just return `null`.

ebma (changes requested):

src/helpers/spacewalk.ts
gonzamontiel: OHH. Okay it was a misunderstanding. Thanks for clarifying that :)

ebma (approved):
Looks good to me now, thanks 👍

PR #140, opened 2023-05-10
Un-hardcode Type column

gonzamontiel:
- Show 'To <chain_name>' instead of 'To Pendulum' in Transfers Type col


gonzamontiel:
fyi this will only be visible on Amplitude when #138 is merged

prayagd:
@gonzamontiel I am not able to see the Amplitude list of transfers. So no way to find out whether the type is changed or not.

gonzamontiel:
> @gonzamontiel I am not able to see the Amplitude list of transfers. So no way to find out whether the type is changed or not.

Regarding that:
> fyi this will only be visible on Amplitude when https://github.com/pendulum-chain/portal/pull/138 is merged

Now, I merged staging to this branch, so you can test it.

prayagd (approved):
yes, i can see the "to Amplitude" type now.

PR #155, opened 2023-05-17
Show vault details in the vault dropdown

gonzamontiel:
- I had to refactor the VaultSelector to use a Dropdown, because items cannot be customized with a simple select (LabelledSector)
- I didn't use the available DropdwonSelector because it assumes the value of the option is a simple string (we want a full custom item)
- **Important**: I used the value of issuedTokens of a measure of vault capacity, but this is probably not correct. Interlay defines vault capacity here: https://docs.interlay.io/#/vault/overview?id=fee-distribution-based-on-btc-capacity. Should we use this approach? We probably need some clarification here from @ebma


gonzamontiel:
BTW, I had to merge #139 to this branch to avoid some errors, so that PR is blocking this one.

ebma:
@gonzamontiel the deploy seems to have failed, can you look into that?

ebma (commented):
I already mentioned it on Slack but I'll say it here again. 

Since we want to show the user how many tokens he can issue with each vault and not how many tokens the vault has issued _so far_ we cannot use `vault.issuedTokens`. Instead we should use the return values of [these two](https://github.com/pendulum-chain/spacewalk/blob/befd6fe8de903407e212902956398df25b50e8f3/pallets/vault-registry/rpc/src/lib.rs#L56C1-L66) RPC functions to receive the amounts of tokens that can be issued/redeemed with the current values of the vault collateral and price info.

gonzamontiel:
> @gonzamontiel the deploy seems to have failed, can you look into that?

Sorry, I was missing one commit from last week. I merge changes from #159 here, so I can show the issuable tokens in the vaults dropdown.

ebma (changes requested):
The dropdown looks nice but as a user I would expect the dropdown to close once I click on an item to select it. Also, the amounts shown are always 0.00 which is probably a bug.

gonzamontiel:
> > The dropdown looks nice but as a user I would expect the dropdown to close once I click on an item to select it. Also, the amounts shown are always 0.00 which is probably a bug.
> 
> We need #159 merged to have this fully working.

It can be functional tested now. I also found annoying that the dropdown does not close automatically when selecting a item, apparently this is intentional in Daisy UI, I'm working on a fix (not that straight forward though)

src/helpers/spacewalk.ts
```diff
@@ -131,7 +131,7 @@ export function currencyToString(currency: SpacewalkPrimitivesCurrencyId, tenant
 function tryConvertCodeToAscii(code: U8aFixed) {
   const ascii = hex_to_ascii(code.toHex());
   if (ascii !== ascii.trim()) {
-    throw Error('Asset code contains invalid space characters');
+    throw Error('Asset has invalid characters');
```
ebma, line 134: Is this a regression from merging?

src/helpers/spacewalk.ts
```diff
@@ -169,3 +169,5 @@ export function estimateRequestCreationTime(
   const now = DateTime.now();
   return now.minus({ seconds: secondsAgo });
 }
+export function calculateVaultCapacity(vault: VaultRegistryVault) {
+}
```
ebma, line 173: Can be removed probably

ebma (approved):
Added two comments for minor issues but LGTM overall 👍

src/helpers/spacewalk.ts
gonzamontiel: bad merge, yes

PR #156, opened 2023-05-18
minor fixes

### LINKED ISSUE #150: Minor fixes
Listing down small fixes and change of text that needs to be done
* Change Amber to 0xAmber in the navigation menu
* Amm to Zenlink AMM in the navigation menu

gonzamontiel:
Closes #150

### DIFF
--- a/src/components/Layout/links.tsx
+++ b/src/components/Layout/links.tsx
@@ -40,7 +40,7 @@ export const links: Links = ({ tenantName }) => [
   },
   {
     link: './amm',
-    title: 'Amm',
+    title: 'Zenlink Amm',
     props: {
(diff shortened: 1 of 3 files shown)


ebma (changes requested):

ebma (approved):
LGTM in theory. But I thought about it and the ticket that this solves doesn't make sense. Why are we showing an item for 'Zenlink AMM' in the first place, if the goal is to use their UI which lives _on their page_. Either this item will become a link to their UI once ready or we should remove it @prayagd.

I'm approving this anyways because it solves the ticket and we can change that later on.

prayagd:
@ebma You are right, this item will become a redirect link to Zenlink UI, the same way we have done for Governance.

PR #166, opened 2023-05-24
Fix xlm dialog

### LINKED ISSUE #161: Update token names on Issue Request message
### Observation
The message on the issue request message currently displays USDC for all tokens. This should display the token being issued.

### Requirement
Update the message to reflect the name of the token being issued.

### LINKED ISSUE #162: Fix missing information on XLM dialog box
### Observation
When creating an issue request for XLM, the information for the stellar transaction is missing.

### Requirement
The dialog box must display the relevant details for the Stellar transaction

ebma:
Closes #161 and closes #162.

Also shows the instruction to include the memo in the 'Pending' dialog on the transfers page. (I think otherwise it's not really clear that you need to include that if a user misses sending it with the first dialog and then tries to re-do it by looking at the transfers page.)

Also fixes an issue with the validation of the issue amount and changes the spacing of the transfers countdown.

### DIFF
--- a/src/pages/bridge/TransferDialog.tsx
+++ b/src/pages/bridge/TransferDialog.tsx
@@ -112,7 +112,7 @@ interface TransferDialogProps {
 
 export function CompletedTransferDialog(props: TransferDialogProps) {
   const { transfer, visible, onClose } = props;
-  const stellarAsset = transfer.original.asset.asStellar.asAlphaNum4.code.toHuman()?.toString();
+  const stellarAsset = currencyToString(transfer.original.asset);
   const content = (
     <>
       <div className="text-md">{`You have received  ${transfer.amount} ${stellarAsset}`}</div>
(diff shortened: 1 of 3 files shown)


src/pages/bridge/TransferDialog.tsx
```diff
@@ -112,7 +112,7 @@ interface TransferDialogProps {
 
 export function CompletedTransferDialog(props: TransferDialogProps) {
   const { transfer, visible, onClose } = props;
-  const stellarAsset = transfer.original.asset.asStellar.asAlphaNum4.code.toHuman()?.toString();
```
gonzamontiel, line 115: What? Weird. I feel like I already fixed this, it should have been some conflict that removed it. But I clearly remember this line was replaced by a conversion to an asset code (I even think you reviewed that, but now I can't find the PR)

gonzamontiel: Or maybe I'm wrong and it wasn't in the dialogs, anyway I have added a function decodeAssetCode() to bridge/helpers.ts that does a similar thing than currencyToString(), we might wanna use the same everywhere

gonzamontiel (approved):
LGTM

src/pages/bridge/TransferDialog.tsx
ebma: Yeah, I remember reading something similar not too long ago haha, but not sure when and where. 
Ah right, we should probably just use one if they serve the same purpose. I think `currencyToString()` is more powerful though since it also tries to map the XCM index to the proper name.

PR #167, opened 2023-05-24
Add orml token balances to Portfolio

### LINKED ISSUE #160: Show bridget assets in the portfolio
Show the amounts of bridged assets that the connected account holds.
Use the following screenshot as a guidance:

image

gonzamontiel:
Closes #160

### DIFF
--- a/src/pages/dashboard/Dashboard.tsx
+++ b/src/pages/dashboard/Dashboard.tsx
@@ -1,8 +1,19 @@
-import { PalletBalancesAccountData } from '@polkadot/types/lookup';
+import { Balance } from '@polkadot/types/interfaces';
+import {
+  OrmlTokensAccountData,
+  PalletBalancesAccountData,
+  SpacewalkPrimitivesCurrencyId,
+} from '@polkadot/types/lookup';
+import _ from 'lodash';
 import { useEffect, useMemo, useState } from 'preact/hooks';
+import { stringify } from 'querystring';
+import { Asset } from 'stellar-sdk';
 import Banner from '../../assets/banner-spacewalk-4x.png';
 import { useGlobalState } from '../../GlobalStateProvider';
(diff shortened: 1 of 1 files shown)


ebma:
@gonzamontiel I changed the logic a bit so that it now uses a typed array which always gets a new entry for each fetched balance (even if it's 0). But otherwise, we might not update the map in case the balance changes to 0 while the user is using the portal. I also changed it to fetch also the balances of the collateral currencies because that might also be interesting to see for users and/or vault operators.

src/pages/dashboard/Dashboard.tsx
```diff
@@ -50,43 +49,42 @@ function Dashboard() {
...
 
-    wrappedCurrencies.forEach((currencyId) => {
+    availableCurrencies.forEach((currencyId) => {
       const walletAddress = ss58Format ? getAddressForFormat(walletAccount.address, ss58Format) : walletAccount.address;
       fetchBridgedTokens(walletAddress, currencyId).then((balance) => {
         setAccountTokenBalances((oldBalances) => {
-          const tokenId = currencyToString(currencyId);
-          if (tokenId && balance) {
-            oldBalances[tokenId] = balance;
+          try {
```
gonzamontiel: I guess this try/catch is not needed, the full currencyToString() is wrapped in a try/catch block already, and right after we are asking for tokenId value making sure it is not null

gonzamontiel (commented):
I added a comment and fixed it. I think it's ready.

gonzamontiel:
> I added a comment and fixed it. I think it's ready.

As I'm the owner, I think @ebma you have the final word approving.

src/pages/dashboard/Dashboard.tsx
gonzamontiel: fixed it.

ebma (approved):
LGTM

PR #168, opened 2023-05-30
Fix wrong parameter used for deriving a vault's Stellar address

ebma:
Fixes [this](https://github.com/pendulum-chain/tasks/issues/65) issue.

I changed the namings in the dialog a bit to make it clearer which address/account ID belongs to which network (Stellar vs parachain).

### DIFF
--- a/src/pages/bridge/TransferDialog.tsx
+++ b/src/pages/bridge/TransferDialog.tsx
@@ -17,6 +17,7 @@ import CancelledDialogIcon from '../../assets/dialog-status-cancelled';
 import WarningDialogIcon from '../../assets/dialog-status-warning';
 import TransferCountdown from '../../components/TransferCountdown';
 import { hexToU8a } from '@polkadot/util';
+import { useVaultRegistryPallet } from '../../hooks/spacewalk/vaultRegistry';
 
 interface BaseTransferDialogProps {
   id: string;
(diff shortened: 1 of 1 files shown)


ebma:
I noticed that I actually still have to change something about 'destination address' vs 'vaults stellar address' because they are not necessarily the same although it's now shown like that.

gonzamontiel:
> I noticed that I actually still have to change something about 'destination address' vs 'vaults stellar address' because they are not necessarily the same although it's now shown like that.

So I should wait for reviewing?

ebma:
> So I should wait for reviewing?


Yes, please @gonzamontiel. I'll ping you again. Sorry for that, only realized that after creating the PR.

ebma:
Ready for review now @gonzamontiel

src/pages/bridge/TransferDialog.tsx
```diff
@@ -11,12 +14,10 @@ import { convertRawHexKeyToPublicKey } from '../../helpers/stellar';
 import { useGlobalState } from '../../GlobalStateProvider';
 import { useSecurityPallet } from '../../hooks/spacewalk/security';
 import { calculateDeadline, currencyToString, deriveShortenedRequestId } from '../../helpers/spacewalk';
-import { useEffect, useMemo, useState } from 'react';
-import { DateTime } from 'luxon';
 import CancelledDialogIcon from '../../assets/dialog-status-cancelled';
 import WarningDialogIcon from '../../assets/dialog-status-warning';
 import TransferCountdown from '../../components/TransferCountdown';
-import { hexToU8a } from '@polkadot/util';
+import { useVaultRegistryPallet } from '../../hooks/spacewalk/vaultRegistry';
```
gonzamontiel, line 20: See my comment below
```suggestion
import { useVaultRegistryPallet } from '../../hooks/spacewalk/vaultRegistry';
import { toTitle } from '../../helpers/string';
```

gonzamontiel (changes requested):
I like the changes! I only added a small suggestion and we can merge

gonzamontiel (approved):
lgtmrn

PR #169, opened 2023-05-30
fix: replace duplicated function and use convertCurrencyToStellarAsset()

### LINKED ISSUE #158: Transform the encoded asset codes of assets with a \0 byte to improve UX
## Problem 
On-chain we now have assets that contain a `\0` byte to pad asset codes that are less than 4 characters. This leads to hex-encoded values being shown in some parts of the UI, see the screenshot.
image

Since it's not easy to see which asset is being referred to here, we should transform this encoding to something human-readable. In this case, we want to see `TZS` instead of `0x545...`. 

## Solution
Before showing the asset code in the UI, we should remove the `\0` bytes.

gonzamontiel:
Closes #158


ebma (approved):
LGTM

PR #171, opened 2023-05-31
Stop loading loop

### LINKED ISSUE #157: Random problem at loading data in portal Tables
### Describe the bug

When changing networks from the chain selector, when accessing table for the first time, and also sometimes when reloading the page on a table, you can notice a double loading of the data. 
The final behavior can vary from different scenarios, but observed ones are:
- Table not updating data at all at changing networks
- Table taking a lot of time to reload data, first showing the wrong data when changing chains
- Table loading first data from a chain, then of the other chain, giving the impression of a doble loading

### Reproduction

1. Go to a page with a table such as `/collators` or `/spacewalk/transfers` on the portal
2. Change the chain from the dropdown, in the top right corner
3. You will likely see data not reloading for a while
4. Now try to reload the page, you might see the double loading issue

### Expected behavior
- The table should load data once
- While loading, a placeholder should be shown
- When changing networks, the whole table needs to be reloaded with the proper data

### Screenshots
TBC

gonzamontiel:
Try to solve the generic problem of multiple loading of the API and consequently of other components.
Also:
Closes #157

### DIFF
--- a/src/GlobalStateProvider.tsx
+++ b/src/GlobalStateProvider.tsx
@@ -11,7 +11,7 @@ import { storageService } from './services/storage/local';
 
 export interface GlobalStateValues {
   tenantName: TenantName;
-  tenantRPC: string;
+  tenantRPC?: string;
   wallet?: WalletAccount;
 }
 export interface GlobalState {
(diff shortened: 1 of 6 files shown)


ebma (approved):
LGTM, good job @gonzamontiel 👍

PR #173, opened 2023-05-31
Fix redeem dialog

### LINKED ISSUE #172: Not able to initiate redeem request.
### Describe the bug

Not able to initiate a redeem request or the bridge button becomes unresponsive

### Reproduction

1. Go to /spacewalk/bridge
2. Click on `Back to stellar`
3.  Enter the amount to be bridged back and the stellar address. 
4. The `Bridge` button would become unresponsive. 
5. If the user now changes the amount, the `Bridge` button becomes responsive again. 

### Expected behavior

After entering the amount and the stellar address, i should be able to create a redeem request.

gonzamontiel:
- Updated confirmation dialog
- Changed Transaction ID to Request ID on the transfers table
- Closes #172

### DIFF
--- a/src/pages/bridge/Issue.tsx
+++ b/src/pages/bridge/Issue.tsx
@@ -113,7 +113,7 @@ function ConfirmationDialog(props: ConfirmationDialogProps): JSX.Element {
 
   const { subscribeActiveBlockNumber } = useSecurityPallet();
   const [activeBlockNumber, setActiveBlockNumber] = useState<number>(0);
-  const [remainingDurationString, setRemainingDurationString] = useState<string>('');
+  const [_, setRemainingDurationString] = useState<string>('');
 
   const totalAmount = useMemo(
     () =>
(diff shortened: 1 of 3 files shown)


ebma (approved):
I pushed a minor change which makes it 'cleaner' now. I now remember that 'watching' the values is important since we use the `stellarAddress` field in other memoized stuff of our `Redeem` component. If I remember correctly, the `getValues()` function would either often return an outdated value ie. lagging behind a bit, or it does not cause a rerender when the form field value changed. I don't remember exactly, but something was weird about it.

Rest looks good, thanks for also taking care of the little things 👍

gonzamontiel:
> I pushed a minor change which makes it 'cleaner' now. I now remember that 'watching' the values is important since we use the `stellarAddress` field in other memoized stuff of our `Redeem` component. If I remember correctly, the `getValues()` function would either often return an outdated value ie. lagging behind a bit, or it does not cause a rerender when the form field value changed. I don't remember exactly, but something was weird about it.
> 
> Rest looks good, thanks for also taking care of the little things +1

It's weird that it worked for you, for me watch() didn't do the trick but maybe I was using it incorrectly. Anyway, merged now :)

PR #181, opened 2023-06-01
170 add update increment rewards option for staking

### LINKED ISSUE #170: Add Update / Increment rewards option for staking
@vadaynujra commented on [Fri May 26 2023](https://github.com/pendulum-chain/tasks/issues/69)

### Requirement / User Story
As a staker (delegator and collator) on Pendulum and Amplitude, I want to be able to increment / update my rewards so that I can claim it, on the portal.

### Task
* Add button that calls the extrinsics incrementCollatorRewards() and incrementDelegatorRewards() first so that the rewards are updated
* Call extrinsic on wallet connect

Example from [Penguin.pe](https://penguin.pe/)
 

---

@vadaynujra commented on [Fri May 26 2023](https://github.com/pendulum-chain/tasks/issues/69#issuecomment-1564473131)

Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/tasks/69?planning-poker) with Zenhub @ebma @gonzamontiel

---

@ebma commented on [Fri May 26 2023](https://github.com/pendulum-chain/tasks/issues/69#issuecomment-1564505229)

I checked and the extrinsics are limited as to who can call them. `incrementCollatorRewards()` can only be called from the account of a collator and `incrementDelegatorRewards()` can only be called from the account of a delegator. 

Thus, the button should actually only call the [incrementDelegatorRewards()](https://github.com/pendulum-chain/pendulum/blob/44f0d6d46da12810c4cef6b9df3d96620360dc5e/pallets/parachain-staking/src/lib.rs#L1763) extrinsic for the currently connected wallet account.
Alternatively we could also try to call both extrinsics when clicking on the button so that we cover both cases and collators as well as delegators can use this functionality to claim their rewards. But then we have to expect one of these extrinsic calls to fail and handle that accordingly.

---

@ebma commented on [Tue May 30 2023](https://github.com/pendulum-chain/tasks/issues/69#issuecomment-1568448744)

Why is this ticket not in the portal repo though @vadaynujra?

---

@gonzamontiel commented on [Tue May 30 2023](https://github.com/pendulum-chain/tasks/issues/69#issuecomment-1568467288)

Hu, I came here to say the same, because I haven't found it on ZenHub and figured out it was mistakenly added to another repo. Is there an easy way to move, I noticed this repo is private, and I cannot move it to a public one :/?

---

@ebma commented on [Tue May 30 2023](https://github.com/pendulum-chain/tasks/issues/69#issuecomment-1568469684)

Yeah there's an option to transfer the issue but I'd like to have @vadaynujra clarify first whether there is a reason for the ticket to live here.

---

@ebma commented on [Tue May 30 2023](https://github.com/pendulum-chain/tasks/issues/69#issuecomment-1568470472)

Oh lol, you're right we can't move it because it's private @gonzamontiel. That's interesting

---

@vadaynujra commented on [Tue May 30 2023](https://github.com/pendulum-chain/tasks/issues/69#issuecomment-1568507896)

@ebma no reason to have added it here. Should be in the portal. Can it be moved?

---

@vadaynujra commented on [Tue May 30 2023](https://github.com/pendulum-chain/tasks/issues/69#issuecomment-1568510041)

Please new create a ticket for the portal and close this @ebma @gonzamontiel ?

gonzamontiel:
- I've randomly added a title to each collator's cell with all the non-null information such as email/twitter/etc. I did this because the user deserves to know, but besides, bc the display name of some of them was a hexadecimal string. Taking into account that others are normally shown, I assumed in those cases there was an error on their end (i.e. when calling the Identity pallet), and the title helps identify the collator name better (see attached screenshot)
Screenshot from 2023-06-01 18 25 05

- I didn't call the extrinsic on wallet connect, because I realize the extrinsic is called by the user, and he/she would have to sign, and pay fees. And it's simply horrible UX to do this without user action.

- Closes #170


src/pages/collators/CollatorRewards.tsx
```diff
@@ -0,0 +1,160 @@
...
+                <h3>{nativeToFormat(userStaking?.amount || '0.00', tokenSymbol)}</h3>
+                <p>My Staking</p>
+              </div>
+              <div className="flex-auto">
+                <h3>{nativeToFormat(userAvailableBalance, tokenSymbol)}</h3>
+                <p>Free balance</p>
+              </div>
+              <div className="flex flex-auto place-content-end">
+                <button className="btn btn-secondary w-full" disabled>
+                  0 {tokenSymbol} Unboarding
```
ebma, line 112: Should this really be 'Unboarding' or 'Unbonding'?

src/pages/collators/CollatorRewards.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+    fetchAvailableBalance().then((balance) => setUserAvailableBalance(balance));
+  }, [api, walletAccount]);
+
+  const updateRewardsExtrinsic = useMemo(() => {
+    if (!api) {
+      return undefined;
+    }
+
+    return createUpdateDelegatorRewardsExtrinsic();
```
ebma, line 59: Can we also update the collator rewards in this function? Maybe we can batch the `incrementDelegatorRewards()` and `incrementCollatorRewards()` extrinsics to one transaction so that the user doesn't have to sign twice. But updating both would be preferrable.

src/pages/collators/CollatorRewards.tsx
```diff
@@ -0,0 +1,160 @@
...
+        </div>
+        <div className="card rounded-lg bg-base-200 w-1/2 ml-4 collators-box">
+          <div className="card-body">
+            <h2 className="card-title">Staking Rewards</h2>
+            <div className="flex flex-row">
+              <div className="flex-initial pt-1 pr-5 pb-0">
+                <RewardsIcon />
+              </div>
+              <div className="flex-auto">
+                <h4>{nativeToFormat(estimatedRewards, tokenSymbol)}</h4>
```
ebma, line 126: The balance here is not updated after clicking on the 'update' button. Reason being that this is kind of cached in the `useStakingPallet()` hook. We should consider changing the function in that hook such that it's possible to fetch the latest estimate of rewards.

ebma (commented):
Looks good to me overall but I added some comments.

src/pages/collators/CollatorRewards.tsx
gonzamontiel: Uhm, so the texts in the designs read 'Unboarding'. But maybe it was a misheard ? @prayagd ?

src/pages/collators/CollatorRewards.tsx
gonzamontiel: I'll take that, will modify it :+1:

src/pages/collators/CollatorRewards.tsx
gonzamontiel: Okay, I'll take a look at this too.

gonzamontiel: So okay, what happens is that useEffect() won't change if no dependency changed, and in this case the dependency is "hidden" because the value changes remotely. So I simply added a function to refresh the estimated rewards.

ebma (approved):
Code looks good, can confirm that updating works now 👍

src/pages/collators/CollatorRewards.tsx
gonzamontiel: I will mark it as solved so we can merge.

PR #186, opened 2023-06-07
add expiry date to local storage hooks

gonzamontiel:
Adds an expiry date option to items in local storage.
- It adds another item to local storage to save the expire date for that item
- I wasn't sure where to put it, so I placed it while rendering the `GlobalStateProvider`
- Not very secure since users can change expire date from storage, so we don't need to rely on this mechanism for anything else than simple usage

### DIFF
--- a/src/hooks/useLocalStorage.ts
+++ b/src/hooks/useLocalStorage.ts
@@ -2,6 +2,8 @@ import { useCallback, useEffect, useMemo, useRef, useState } from 'preact/compat
 import { debounce } from '../helpers/function';
 import { storageService } from '../services/storage/local';
 import { Storage } from '../services/storage/types';
+import { DateTime } from 'luxon';
+import { storageKeys } from '../constants/localStorage';
 
 export type UseLocalStorageProps<T> = {
   /** Storage key */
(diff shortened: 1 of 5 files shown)


src/hooks/useLocalStorage.ts
```diff
@@ -60,6 +65,13 @@ export const useLocalStorage = <T>({
     (value) => {
       storageSet(key, value);
       setState(value);
+      if (expire) {
+        const expiryDate = DateTime.now().plus({ seconds: expire });
+        const expiryKey = `${key}${storageKeys.EXPIRY_DATE}`;
+        storageSet(expiryKey, expiryDate);
+        debounce(storageService.remove, expire * 1000)(key);
```
ebma, line 72: While not added as part of this PR, I think the `debounce` function is confusing. Because if you compare it to the [debounce](https://docs-lodash.com/v4/debounce/) function of lodash, they are quite different actually. And since the function is actually just a more convenient implementation to create a timer, I would rename it to indicate such. Maybe rename it to `createTimer()` or `periodicCall()` or whatever.

ebma (approved):
The logic looks good overall, I just commented on something unrelated.

src/hooks/useLocalStorage.ts
gonzamontiel: Once this conversation is resolved, please fell free to merge :)

nejcm: So debounce is debouncing the writing to the storage. This is used cause we write data to localstorage on input change which can be quite frequent so I am debouncing this for 1s or so.

ebma: Ahhhh now I get it. I missed the point of the `let timer ...` outside of the timeout function. Okay, then I guess that works.

PR #191, opened 2023-06-14
fix: lint issues

nejcm:

PR #192, opened 2023-06-14
style: responsive UI

### LINKED ISSUE #94: Add responsiveness on portal for mobile devices
### Context / Issue
The current portal UI on mobile devices is not acceptable as a lot of features do not work.

### Requirement / User Story
As a mobile user, I want to the portal website to be responsive to my device, so that I can use all the features of the portal.

### Solution / Tasks

As an interim solution, since we don't have any designs for this, we first try to make every feature 'usable' meaning that users can access/see all information and can interact with every button, even though the resulting user experience might still not be ideal. To do that:
- [ ] Introduce responsive design in the best possible way, without requiring new designs.
This means that the existing UI elements are probably just re-arranged and re-scaled if needed. 
- [ ] To test the responsiveness of the current version of the portal.

nejcm:
Closes #94.


nejcm:
This PR is dependent on the https://github.com/pendulum-chain/portal/pull/191 that is why there are so many file changes.

ebma:
We might want to change the color of the hamburger icon on the Amplitude theme, it's really hard to see.

image

Also, the spacewalk UI does not work too well on e.g. 'iPhone 12 Pro'. Probably because the dialogs have a min-width. 
image

Maybe we could also add a click-away listener that closes the sidebar again? Because at the moment I have to click on the hamburger icon again to hide it which might be counter-intuitive.


The rest looks quite good though 👍

ebma:
Also @nejcm could you maybe add a brief description to your PRs when it's not obvious what changed? For the linting, it's a no-brainer but here it would be nice to see for a reviewer what you intended to change and then compare it to the actual code changes.

nejcm:
@ebma Done

ebma:
Wow, it looks and works really amazing now, great job @nejcm! However, the sidebar is somewhat annoying because I have to manually close it, can't we add a click-away listener? Or instead, we could also make it close automatically if a sidebar item is clicked that changes the navigation of the page.

nejcm:
> Wow, it looks and works really amazing now, great job @nejcm! However, the sidebar is somewhat annoying because I have to manually close it, can't we add a click-away listener? Or instead, we could also make it close automatically if a sidebar item is clicked that changes the navigation of the page.

Done

ebma (approved):
I love it, great job @nejcm

PR #195, opened 2023-06-21
feat: wallet connect

nejcm:
WIP


src/components/Wallet/WalletConnect/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+export type WalletConnectProps = {
+  setWalletAccount: GlobalState['setWalletAccount'];
+};
+
+const wcParams = {
+  requiredNamespaces: {
+    polkadot: {
+      methods: ['polkadot_signTransaction', 'polkadot_signMessage'],
+      chains: [
+        'polkadot:91b171bb158e2d3848fa23a9f1c25182', // polkadot
```
ebma, line 18:
```suggestion
        'polkadot:67221cd96c1551b72d55f65164d6a39f', // foucoco,
        'polkadot:cceae7f3b9947cdb67369c026ef78efa', // amplitude
        'polkadot:5d3c298622d5634ed019bf61ea4b7165', // pendulum
        
```

ebma (commented):
Edit: We can use the Nova Wallet mobile app to test the walletconnect communication.
I did that and it doesn't seem to work. The console shows lots of websocket errors. @nejcm can you look into this maybe?

nejcm:
> Edit: We can use the Nova Wallet mobile app to test the walletconnect communication. I did that and it doesn't seem to work. The console shows lots of websocket errors. @nejcm can you look into this maybe?

Fixed. Some domain exceptions needed to be added to wallet connect dash

prayagd:
@nejc Moving this to development and removing the WIP label as it means something is incomplete and ticket is yet to prepared.

nejcm:
@ebma @prayagd This is ready for final testing and merge. I added the persistence to wallet connect.

ebma (changes requested):
When disconnecting from the wallet (using the disconnect button in the top-right hand corner) and refreshing the page, it will still load the previously disconnected account.

It seems that there are also more inconsistencies in regard to the local-storage. When you first connect with an account from polkadot.js extension, and then disconnect from that, it does not disconnect immediately. Or at least it's not reflected properly. You have to disconnect twice. But if after you used a polkadot.js account, you decide to use walletconnect, the UI will show this account as imported from polkadot.js as well. And it's not even the same account. Also, refreshing the page sometimes imports a polkadot.js account and sometimes the walletconnect account.

@nejcm can you look into this again? Seems like some local storage keys are not properly cleared and the Wallet button is not re-rendered properly. 

Last but not least, there is some weird issue that happens when you play around with the wallet choices shown in the dialog. If a user clicks on one of these items, a certain item will be placed in local storage. When the user now needs to sign a transaction, the browser will almost immediately redirect to that other wallet. This is super bad, because we want the user to see the request in their smartphone wallet. To circumvent this behavior, I think our only good choice is to always clear that respective local storage key so that it is unset at all times.
image
image

nejcm:
Managed to fix the storage issue but cant recreate the wallet connect issue with apps. That is the internal functionality of the library so we will have to do some "hacky" stuff to work around that.

ebma:
@nejcm can you resolve the conflicts? Sorry for that, I just merged the other PR.

ebma (approved):
Although it's not perfect yet, we'll merge it already.

Issue #178, opened 2023-06-01
Redirect to website when clicked on the logo


prayagd:
As a portal user, on clicking on the logo of Pendulum/Amplitude the respective website should open in new tab. 

## Acceptance criteria
* Open the Pendulum/Amplitude website in new tab when clicked on the logo. 
* Pendulum - https://pendulumchain.org/
* Amplitude - https://pendulumchain.org/amplitude

PR #199, opened 2023-06-27
feat: redirect on logo click  #178

nejcm:


ebma (commented):
It does what ticket #178 says it should do. But why are we opening the page in a new tab? Why are we not just navigating back to the dashboard _in the same tab_ when I click on the logo? As a user, I find that counter-intuitive and annoying. @prayagd is there a reason we use a new tab?

src/components/Layout/index.tsx
```diff
@@ -63,13 +63,16 @@ export default function Layout(): JSX.Element | null {
...
-            alt=""
-            style={isPendulum ? {} : { marginTop: 20, marginBottom: 30, marginLeft: 30 }}
-          />
-          {isTestnet && <div className="foucoco-tag">Foucoco testnet</div>}
+          <div
+            style={
+              isPendulum ? { marginLeft: 20, marginBottom: 20 } : { marginTop: 20, marginBottom: 30, marginLeft: 30 }
+            }
+          >
+            <Link to={`/${network}/dashboard`} target="_blank">
```
ebma, line 71: This needs to change to either https://pendulumchain.org/ when the network is Pendulum or https://pendulumchain.org/amplitude when the network is Foucoco/Amplitude.

ebma (changes requested):
Okay got it @prayagd. Thanks for clarifying. Now that you mention it, the current implementation does not actually link to the websites but just navigates to the dashboard. I requested changes from @nejcm

ebma (approved):
LGTM

Issue #204, opened 2023-07-03
Fix persistence issues with WalletConnect


ebma:
The account persistence is not working properly. You can reproduce the issue by following these steps: 
1. Clear your localStorage for the portal page. 
3. Connect to any account that is imported into your polkadot.js extension. 
4. Disconnect from that account again. 
5. Connect to an account using wallet connect. 
6. Refresh the page. 

You will notice that the account that is loaded from storage is different from the account you connected to using WalletConnect.

PR #209, opened 2023-07-09
fix: wallet connection issues #204

nejcm:


src/GlobalStateProvider.tsx
```diff
@@ -3,13 +3,13 @@ import { createContext } from 'preact';
 import { useCallback, useContext, useEffect, useMemo, useRef, useState } from 'preact/compat';
 import { useLocation } from 'react-router-dom';
 import { config } from './config';
+import { chainIds } from './config/walletConnect';
```
ebma, line 6: Do we need this import?

ebma (approved):
LGTM overall

src/GlobalStateProvider.tsx
gonzamontiel: I see this is indeed used in [line 97](https://github.com/pendulum-chain/portal/blob/ad6735de6f05dcecb2cb5c3d9c730775d8d6b232/src/GlobalStateProvider.tsx#L97) in the same file, I think we can resolve this conv and merge?

PR #211, opened 2023-07-14
feat: token approval

nejcm:


nejcm:
We could potentially merge this into staging as it only touches nabla even though it's incomplete until the contracts start working

gonzamontiel (approved):
Overall approved.
There are still some error handling to improve (as discussed) and code that could be reused and it's duplicated (for example in the Swap modals).
Some components could be extracted and we could actually used in other parts (like the To and From inputs, the asset selector, etc.).
I appreciate the overall effort, thanks.
Also, thanks for addressing the naming changes I requested.

PR #214, opened 2023-07-25
174 dont initiate issue or redeem requests for invalid values

### LINKED ISSUE #174: Dont initiate issue or redeem requests for invalid values
### Describe the bug

If the user enters 0 or value exceeding max issuable or redeemable value the request should not be initiated

### Reproduction

1. Go to /spacewalk/bridge
2. Click on to Amplitude
3. Enter 0 in the input field and click `Bridge`
4. The issue requests get initiated. 
5. Same if the user enters value greater than `Max issuable`. The issue requests get initiated

### Expected behavior

The user should be not be able to initiate a issue or redeem request for value 0 and values greater than the vault capacity. 
Warning should be show saying invalid values if the user tries to click on bridge with these values.

gonzamontiel:
Closes #174

### DIFF
--- a/src/pages/bridge/Issue.tsx
+++ b/src/pages/bridge/Issue.tsx
@@ -382,7 +382,7 @@ function Issue(props: IssueProps): JSX.Element {
         onClose={() => setConfirmationDialogVisible(false)}
       />
       <div className="w-full">
-        <form className="px-5 flex flex-col" onSubmit={handleSubmit(submitRequestIssueExtrinsic)}>
+        <form className="px-5 flex flex-col">
           <div className="flex flex-col xs:flex-row xs:items-center gap-1">
             <Controller
               control={control}
@@ -394,6 +394,9 @@ function Issue(props: IssueProps): JSX.Element {
                   }
                   if (parseFloat(value) > parseFloat(selectedVault?.issuableTokens?.toString() || '0')) {
                     return 'Amount is higher than the vault can issue.';
+                  }
+                  if (parseFloat(value) <= 0) {
+                    return 'Amount should be a positive number.';
                   }
                 },
               }}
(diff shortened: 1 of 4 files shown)


ebma (changes requested):
Is there a reason why you moved the onSubmit function to the button itself @gonzamontiel?

The changes introduce a bug. To reproduce:
Without having a connected wallet account, if you click on the "Connect to Wallet" shown below the form, the page will reload and you cannot connect. Meaning, the "Connect to Wallet" button is broken somehow.

Can confirm that the form validation works though.

gonzamontiel:
> The changes introduce a bug. To reproduce: Without having a connected wallet account, if you click on the "Connect to Wallet" shown below the form, the page will reload and you cannot connect. Meaning, the "Connect to Wallet" button is broken somehow.

So, I didn't introduce any bug, I just made a pre-existing one more evident.
What happens is that our WalletConnect widget uses the Button component, and unfortunately, they are of type "submit" by default (check the most voted answer [here](https://stackoverflow.com/questions/932653/how-to-prevent-buttons-from-submitting-forms)). So the form was being submitted involuntarily.
Actually you can check the behavior yourself in [staging](https://staging--rococo-souffle-a625f5.netlify.app/amplitude/spacewalk/bridge) , you'll see that while the OpenWallet dialog is shown, a transaction is sent and an error appears ("no wallet account selected").
I fixed it by adding _type="button"_ to those.

src/pages/bridge/Issue.tsx
```diff
@@ -395,6 +395,9 @@ function Issue(props: IssueProps): JSX.Element {
                   if (parseFloat(value) > parseFloat(selectedVault?.issuableTokens?.toString() || '0')) {
```
ebma, line 395: (Although not related to this PR and a bug that already existed before)
I suppose we need to downscale this amount to 'whole units' ie. divide it by our decimals (12). 
You can see that it's not working as expected by testing with the vaults registered on Foucoco, the entered amount is higher than 700, but the validation is not triggered.
image

ebma (changes requested):
Nice, can confirm that the connect wallet button is fixed. But I found another bug.

src/pages/bridge/Issue.tsx
gonzamontiel: You are right, nice catch. I fixed it.

PR #215, opened 2023-07-25
Fix versions & runtime in navbar

### LINKED ISSUE #144: Fix Dapp, Runtime and Network
Currently the DApp, Runtime and Network are confusing to the enduser. We are not aware what information is fetched here. So writing this ticket to make minor fixes. 

## Requirement
* Remove Dapp
* Keep Runtime
* Change Network to Block no.

gonzamontiel:
Closes #144


ebma (approved):
LGTM
Would be better of course if the last block number actually updated with every new block but I think that's too much effort for such a minor detail.

PR #228, opened 2023-08-08
177 only show transfers for connected account

### LINKED ISSUE #177: Only show transfers for connected account
### Observation
Currently all users can view all bridge requests under the Transfers tab.

### Requirement
Only transfers from the connected accounts should be displayed.
When no account is connected, no transfers are displayed.

gonzamontiel:
Closes #177


src/helpers/spacewalk.ts
```diff
@@ -128,6 +128,13 @@ export function currencyToString(currency: SpacewalkPrimitivesCurrencyId, tenant
   }
 }
 
+export function currencyToStellarAssetCode(currency: SpacewalkPrimitivesCurrencyId, tenant: TenantName = TenantName.Pendulum) {
+  if (!currency.isStellar) {
+    throw Error('Not an Stellar asset');
```
ebma, line 133: This function throws an error that is not caught by any calling function of the transfer dialogs. If you test this on Amplitude, this error is thrown a lot and since it's an uncaught error, the application stops working appropriately (ie. you cannot close any dialogs or do anything else).
I recommend we wrap this in a try-catch block like above and return `'Unknown'` or `null`.

ebma: Also, as defined [here](https://github.com/pendulum-chain/portal/blob/de0fe699d2b70882829aff272d30a909bfe1491c/src/pages/bridge/TransfersColumns.tsx#L19-L20), the `asset` is a string and not of type `SpacewalkPrimitivesCurrencyId` which is probably the reason why this function errors all the time.

src/pages/bridge/Transfers.tsx
```diff
@@ -53,14 +53,17 @@ function Transfers(): JSX.Element {
       const entries: TTransfer[] = [];
 
       issueEntries.forEach((e) => {
+        if (!walletAccount || e.request.requester.eq(walletAccount?.address)) {
```
ebma, line 56: The condition is wrong and will only show the requests that the account did not create.

```suggestion
        if (!walletAccount || !e.request.requester.eq(walletAccount?.address)) {
```

src/pages/bridge/Transfers.tsx
```diff
@@ -73,6 +76,9 @@ function Transfers(): JSX.Element {
       });
 
       redeemEntries.forEach((e) => {
+        if (!walletAccount || e.request.redeemer.eq(walletAccount?.address)) {
```
ebma, line 79:
```suggestion
        if (!walletAccount || !e.request.redeemer.eq(walletAccount?.address)) {
```

ebma (changes requested):

src/helpers/spacewalk.ts
gonzamontiel: Thanks for the catch. Couple of things here:
1. `currencyToStellarAssetCode()` takes the same signature as `currencyToString()`, that's why it should receive a`SpacewalkPrimitivesCurrencyId`, and that was already working good afaik (i.e. I just replaced the uses of `currencyToString` with `currencyToStellarAssetCode`, the expected behavior was to show only the asset code, instead of the whole `code:issuer`)
2. I didn't realize that there's another existing function called `convertCurrencyToStellarAsset()` which already does the check, even though it doesn't throw an error, which was a bad decision after all. I will use that one and check why is failing so much.

ebma (changes requested):
The code to solve the original issue ticket looks good now.

But the changes for the redeem dialog are wrong. If it timed out and is still pending, then we cannot just say it's cancelled. Because it's not. The user now can take actions for retrying or reimbursing. Instead it should show the following dialog, that I took from the [figma](https://www.figma.com/file/5up5sghcttBogehUq4255i/App?type=design&node-id=419-616&mode=design&t=h9I1eQxFyNay72nU-0).
image

Thus, I would like these changes to be removed, we should create a follow-up ticket to show proper pending redeem dialogs @prayagd.

prayagd:
@ebma @gonzamontiel Yes there would be tickets for updating the dailog boxes.

PR #232, opened 2023-08-16
feat: indexer

nejcm:
This implements indexer queries and hooks for nabla


src/components/Asset/Selector/Modal/index.tsx
```diff
@@ -35,13 +35,13 @@ const AssetList = ({ assets, onSelect, selected }: AssetListProps): JSX.Element
...
             variant="secondary"
-            key={token.address}
+            key={token.id}
             onClick={() => onSelect(token)}
             className="w-full items-center justify-start gap-4 px-3 py-1 border-0 bg-[rgba(0,0,0,.2)] text-left hover:opacity-80 dark:bg-[rgba(255,255,255,.12)]"
           >
             <span className="relative">
-              <Avatar size="xs" letters={token.symbol} src={token.logoURI} shape="circle" className="text-xs" />
-              {selected == token.address && (
+              <Avatar size="xs" letters={token.symbol} /* src={token.logoURI} */ shape="circle" className="text-xs" />
```
ebma, line 43: What are we going to do about the token logo?

src/components/Pools/Swap/columns.tsx
```diff
@@ -48,7 +48,7 @@ export const coverageColumn: ColumnDef<SwapPoolColumn> = {
 export const aprColumn: ColumnDef<SwapPoolColumn> = {
   header: 'APR',
   accessorKey: 'apr',
-  accessorFn: (row) => row.apr?.toString(),
+  accessorFn: (row) => row.reserves?.toString(), // TODO: get apr
```
ebma, line 51: Is this something that we calculate or that the indexer needs to provide?

src/components/Swap/Approval/index.tsx
```diff
@@ -10,7 +10,7 @@ export interface ApprovalProps {
 }
 
 const ApprovalSubmit = ({ token }: ApprovalProps): JSX.Element | null => {
-  const { router } = useGetTenantData(nablaConfig) || {};
+  const { router } = useGetTenantData(nablaConfig)[0] || {};
```
ebma, line 13: Can we maybe make this more expressive?

ebma: I see that it's convenient to use it this way, but it's hard to understand from the naming etc. why you have to access the first part of an array. Maybe we just remove the `tenantName` from the array and only return the data of the current tenant here? The `tenantName` probably is not required most of the time.

ebma (approved):
Added some comments but the code looks good overall 👍

src/components/Asset/Selector/Modal/index.tsx
nejcm: So on nabla we have a constant with images from coingecko. Maybe there is a better way

src/components/Pools/Swap/columns.tsx
nejcm: This is something we did not resolve yet on nabla side either. We have a meeting tmr

src/components/Swap/Approval/index.tsx
nejcm: Sure thing

src/components/Asset/Selector/Modal/index.tsx
ebma: So you have a hard-coded mapping of token symbol to logo?

src/components/Swap/Approval/index.tsx
nejcm: Updated

ebma (approved):
Looks good, thanks

src/components/Asset/Selector/Modal/index.tsx
nejcm: No logo currently. So in that avatar if the image is missing it defaults to showing the symbol. This is useful even once we have images is one does not load or is missing.

nejcm:
@ebma I dont have the perms to merge so you can do it.

PR #233, opened 2023-08-16
feat: contract hooks

nejcm:
@ebma I dried reimplementing the contract gas calc based on that article and it is still failing. Maybe we should try adding 10% on top of that value but at this point I am not sure how to make this work. I am open to ideas. You can try running this branch and checking useContractWrite.
Is it possible there is an issue with deployed contracts. Is it possible to test the contract method somewhere else (online) to see if it works


ebma:
So I played around with it a bit and I would guess the issue is that you schedule two transactions right after each other. I tested it with the 'deposit' functionality of the swap pools page. And when I sign and submit the first transaction (which is the approval function), I am immediately asked to sign another transaction. 

One account cannot submit multiple transactions per block. Or to be precise, you can do that but they might clash because they will have the same priority and only one will get executed. So if you want to do the approval and deposit extrinsic right after each other, you have to bundle them in another transaction basically. You can batch extrinsic calls with the `batchAll` extrinsic. The usage is quite straightforward. For example you could
```
 const transactions = [
       erc20contract.tx.call(..approve.),
       erc20Contract.tx.call(...deposit...)
    ];

    const call = api.tx.utility.batchAll(transactions);

```

That's just some pseudo-code that shows how you can use that to combine other extrinsic calls. I'm not sure if there might be problems if you approve and deposit in the same block but that's to be seen. Alternatively you could of course also approve, wait for the block finalization, then deposit.

But I think that's one part of the problem.

nejcm:
> So I played around with it a bit and I would guess the issue is that you schedule two transactions right after each other. I tested it with the 'deposit' functionality of the swap pools page. And when I sign and submit the first transaction (which is the approval function), I am immediately asked to sign another transaction.
> 
> One account cannot submit multiple transactions per block. Or to be precise, you can do that but they might clash because they will have the same priority and only one will get executed. So if you want to do the approval and deposit extrinsic right after each other, you have to bundle them in another transaction basically. You can batch extrinsic calls with the `batchAll` extrinsic. The usage is quite straightforward. For example you could
> 
> ```
>  const transactions = [
>        erc20contract.tx.call(..approve.),
>        erc20Contract.tx.call(...deposit...)
>     ];
> 
>     const call = api.tx.utility.batchAll(transactions);
> ```
> 
> That's just some pseudo-code that shows how you can use that to combine other extrinsic calls. I'm not sure if there might be problems if you approve and deposit in the same block but that's to be seen. Alternatively you could of course also approve, wait for the block finalization, then deposit.
> 
> But I think that's one part of the problem.

I see. But each transaction should be awaited. So approval has to complete before the other transaction can be executed otherwise the tokens spending would not be sufficient. The transaction anyway fails. Eg. You approve tokens and than execute the deposit much later. It still fails. We can jump on a quick call.

nejcm:
@ebma @gonzamontiel Ok I believe this is at the point where we can merge it. Now there are still issues with contract execution but those can be resolved in another PR.
Issues are with the balance checking and withdrawing from pools.

src/shared/useContractWrite.ts
```diff
@@ -47,20 +47,14 @@ export const useContractWrite = <TAbi extends Abi | Record<string, unknown>>({
...
-        walletAddress,
-        { storageDepositLimit: null, gasLimit: -1 },
-        ...fnArgs,
-      );
-      contractOptions = {
-        gasLimit: api.registry.createType('WeightV2', gasRequired) as WeightV2,
-        storageDepositLimit: null,
-      };
-    }
+    console.log(fnArgs);
```
ebma, line 50: This log can be removed probably.

PR #236, opened 2023-08-24
fix: release

nejcm:
You can merge it asap. This should fix the release. I believe the token does not have some permissions needed so it fails to fetch with depth: 0. Hopefully it did not expire.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: pr
+name: ci
 on:
   pull_request:
     branches: [main, staging, develop]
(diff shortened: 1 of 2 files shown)


ebma (approved):

ebma:
The release job still fails. I tried to follow the steps in [this](https://github.com/semantic-release/semantic-release/issues/2636) issue ticket and created a personal access token since I have admin permissions and should be able to bypass the branch protection rules but to no avail. We should investigate this issue again.

nejcm:
> The release job still fails. I tried to follow the steps in [this](https://github.com/semantic-release/semantic-release/issues/2636) issue ticket and created a personal access token since I have admin permissions and should be able to bypass the branch protection rules but to no avail. We should investigate this issue again.

Yeah I believe the fetch depth is needed for the release action. So reverting in another PR. Then we are back to the initial issue which I believe it is a token problem. The GH_TOKEN need all repo permissions

PR #240, opened 2023-08-31
145 Show connected accounts token wise breakdown on the dashboard

gonzamontiel:
- Added a hook with different providers (MEXC and DIA) for getting token prices
- The hook already creates a prices cache but they can be fetched by token
- The alignment of the total value will be to the end of the box (as discussed with product) since align it to the last column is not easy since columns are aligned dynamically depending on screen size.
-  Icons (sent by Prayag) were modified to fit dimensions and all are rounded now for consistency

~[Missing] The MEXC API throws a CORS policy error, I tried some workaorunds but still the same. My guess is that we need to get an API Key to properly access the API, even though is weird because the price can be accessed from the browser.~

UPDATE
- We no longer use MEXC for PEN Token, we use DIA. Thus, Mexc provider was removed.


ebma (commented):
I fixed a bug showing DOT on Amplitude and Foucoco.
Comparing the current version to the wireframe, there are a few things I noticed.
- the promo banner should take up half the screen but takes 2/3
- the token name and icon are bigger than they should
- the values in the 'price' column should be prefixed with a dollar sign
- the amounts should have a ',' delimiter for every three places before the decimal point
image

gonzamontiel:
> I fixed a bug showing DOT on Amplitude and Foucoco. Comparing the current version to the wireframe, there are a few things I noticed.
>
thanks!
 
> * the promo banner should take up half the screen but takes 2/3
I think the design doesn't take into account different screen sizes, we can discuss this with Marina, but it is similar to what I said about the total USD value alignment in respect to the table size. 

> * the token name and icon are bigger than they should
Reason:  I modified all the PNG to take all the width/height of the canvas. Otherwise, some of the icons looked bigger than the others. While doing this, I made everyone look bigger, I'll adjust it ;)

> * the values in the 'price' column should be prefixed with a dollar sign
fixing it

> * the amounts should have a ',' delimiter for every three places before the decimal point
I'd say we need consistency with this all over the app. So we can discuss today and implement it app-wise
 
> image

gonzamontiel:
I addressed the comments above. However, I'm still struggling with the CORS policy error.
Screenshot from 2023-09-05 12 39 24

To try to get rid of this, I see two options:
1. Use [cors-anywhere](https://github.com/Rob--W/cors-anywhere), create our own CORS proxy (not 3rd party) and use it.
2. Use MEXC [SDK](https://github.com/mxcdevelop/mexc-api-sdk/),  which for me was overkilling but I guess somehow will fix the problem

prayagd:
@gonzamontiel DIA now has the PEN price feed ready, shared with you on #portal channel

gonzamontiel:
> @gonzamontiel DIA now has the PEN price feed ready, shared with you on #portal channel

Then the PR is ready :)

ebma (approved):

PR #251, opened 2023-09-12
fix: WalletConnect issues with non-polkadot wallets

### LINKED ISSUE #213: Fix WalletConnect connection possible with non-polkadot wallets
As reported by @vadaynujra, at the moment it's possible to connect to the portal with the WalletConnect feature of wallets that are not holding/supporting Substrate accounts. 

This was tested and reproduced with the LOBSTR wallet app on an Android phone. The user was able to scan the QR code shown in the portal's WalletConnect dialog, which then after reloading the page lead to the following account being shown.
image

We should generally try to limit the connections to wallets that support Substrate accounts.

gonzamontiel:
Closes #213


Sharqiewicz:
The issue is being resolved in #498

PR #275, opened 2023-10-18
release staging to master

gonzamontiel:


prayagd (commented):
Just one i see Nabla menu item is activated on Amplitude network, but we dont want to. One better thing would be to not Nabla menu on Amplitude network at all, as Nabla wont launch on Amplitude but only on Pendulum and Foucoco

PR #278, opened 2023-10-30
fix: Show vaults for selected asset on Bridge Settings

### LINKED ISSUE #272: Manual vault selection always showing vaults for USDC
### Describe the bug

The manual vault selection drop-down always only shows the vaults for USDC, no matter which asset I select for bridging.

### Reproduction

1. Go to https://staging--rococo-souffle-a625f5.netlify.app/amplitude/spacewalk/bridge
2. Select an asset other than 'USDC' next to the amount text field
3. Click on the cog wheel to show the manual vault selection 
4. Try to select a different vault manually

### Expected behavior

The selector should show only the vaults that have the selected asset as wrapped asset.

### Screenshots

image

Notice how in the amount field, 'TZS' is selected but the drop-down still only shows vaults for USDC.

gonzamontiel:
Closes #272

### DIFF
--- a/src/hooks/spacewalk/useBridgeSettings.ts
+++ b/src/hooks/spacewalk/useBridgeSettings.ts
@@ -1,8 +1,9 @@
 import _ from 'lodash';
-import { StateUpdater, useEffect, useMemo, useState } from 'react';
+import React, { StateUpdater, useEffect, useMemo, useState } from 'react';
 import { Asset } from 'stellar-sdk';
 import { convertCurrencyToStellarAsset } from '../../helpers/spacewalk';
 import { stringifyStellarAsset } from '../../helpers/stellar';
+import { BridgeContext } from '../../pages/bridge';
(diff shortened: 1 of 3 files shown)


src/hooks/spacewalk/useBridgeSettings.ts
```diff
@@ -1,8 +1,9 @@
 import _ from 'lodash';
-import { StateUpdater, useEffect, useMemo, useState } from 'react';
+import React, { StateUpdater, useEffect, useMemo, useState } from 'react';
```
ebma, line 2: Should we import this from `preact`? I forgot if this makes a difference.

ebma (approved):
Works and looks good to me

src/hooks/spacewalk/useBridgeSettings.ts
gonzamontiel: I think we could be consistent with the imports (but they are not across the app in many places).  I don't think makes a big difference tbh.

PR #279, opened 2023-10-31
feat: nabla UI and functionality

nejcm:


ebma:
The navigation seems broken at the moment. When I am on Amplitude or Foucoco and click on anything Nabla related in the side bar, I get redirected to the 'Dashboard' page on Pendulum instead. However, we should disable the Nabla tab on Pendulum, hide it completely on Amplitude, and show it and fix the navigation on Foucoco.

nejcm:
> The navigation seems broken at the moment. When I am on Amplitude or Foucoco and click on anything Nabla related in the side bar, I get redirected to the 'Dashboard' page on Pendulum instead. However, we should disable the Nabla tab on Pendulum, hide it completely on Amplitude, and show it and fix the navigation on Foucoco.

Yeah I committed the fix but as mentioned nabla is disabled on prod and I believe the PRs get deployed with prod flag on. So there are two ways of around this. One is testing locally the branch and the other is adding a new env variable 
`VITE_ENVIRONMENT=staging`

src/app.tsx
```diff
@@ -40,12 +40,12 @@ export function App() {
             <Route path="bridge" element={Bridge} />
             <Route path="transfers" element={TransfersPage} />
           </Route>
-          <Route path="nabla" Component={() => <AppsProvider app="nabla" />}>
+          <Route path="nabla" Component={() => <AppsProvider environment={['staging', 'development']} app="nabla" />}>
```
nejcm, line 43: @ebma This is the filter here

nejcm:
> > Yeah I committed the fix but as mentioned nabla is disabled on prod and I believe the PRs get deployed with prod flag on.
> 
> I'm not sure I understand. I think there is a problem either way. I was not necessarily complaining about not being able to see the Nabla features in the preview (ie. not being able to see the swap view) but the navigation is confusing. If what you are saying is true and Nabla is disabled in the preview because it's deployed as 'prod', then it is not properly disabled because I can still click on the Nabla nav items on Amplitude and Foucoco. However as I said the expected behaviour would be to disable the Nabla tab on Pendulum, hide it completely on Amplitude, and show it and fix the navigation on Foucoco ie don't make it redirect.

Correct. Let me take a look.

nejcm:
@ebma Ok updated the navigation to better handle visibility of items

src/config/apps/nabla.ts
```diff
@@ -15,10 +15,13 @@ export type NablaConfig = AppConfig &
 
 export const nablaConfig: NablaConfig = {
   tenants: [TenantName.Foucoco],
+  environment: ['staging', 'development'],
```
ebma, line 18: This indicates for which environments Nabla should be shown? Or what is it for?

src/shared/useContract.ts
```diff
@@ -30,19 +41,12 @@ export const useContract = <T extends Abi | Record<string, unknown>>(
...
-                    proofSize: '1000000',
-                  }),
-                  storageDepositLimit: null,
-                };
+          const opts = getOptions(options, api);
           const response = await contract.query[method](owner, opts, ...(args || []));
           if (!response?.result?.isOk || response?.output === undefined) throw response;
-          return response;
+          // ? TODO: maybe not ideal to cache only output
+          // caching the whole object causes the output to be converted to hex string
```
ebma, line 48: Does this lead to issues?

ebma (commented):
When not connected to a wallet account, the balance fields in the backstop pool and swap pool view show the loading indicator although nothing is actually loading. We should maybe show a warning message or button that the user needs to connect to a wallet account first.

src/config/apps/nabla.ts
nejcm: Exactly. Can add a comment there.

src/shared/useContract.ts
nejcm: So there is some more investigation I wish to do on this but the problem is that the value received when being cache gets serialized with JSON.stringify fn which converts that value into a hex string. That causes issues when reading value from the cache and trying to convert it back because the result!=cached result.  So this implementation works but would like to put some more time into this to find out if there is a better way.

ebma: Hmm okay, well I'm fine if we postpone the investigation to another PR if the current solution works for now.

ebma (approved):
The changes look good to me overall

src/shared/useContract.ts
nejcm: Yes the solution works and is satisfactory to the functionality that is needed. But because the hooks is reusable it might be useful to cache the whole result and not just the output part.

nejcm:
@ebma I dont have the perms to merge this so will leave it up to you to decide on it.

PR #281, opened 2023-10-31
fix: release

nejcm:


.github/workflows/release.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
-          token: ${{ secrets.GH_TOKEN }}
```
gonzamontiel, line 24: this is weird, our variable is called GH_TOKEN

release.config.cjs
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-  branches: ['staging'],
+  branches: ['staging', 'develop'],
```
gonzamontiel, line 2: this won't make any difference

gonzamontiel (approved):
I'll approve this to test if this fixes the GH_TOKEN problem on staging.

PR #291, opened 2023-11-07
feat: add info to the redeem from backstop pool UI

nejcm:


src/components/nabla/Pools/Backstop/index.tsx
```diff
@@ -12,7 +12,7 @@ const BackstopPoolsBody = (): JSX.Element | null => {
   const { data, isLoading } = useBackstopPools();
 
   if (isLoading) return <Skeleton className="bg-neutral-200 h-48 w-full" />;
-  const pool = data?.[0];
+  const pool = data?.[data.length - 1];
```
ebma, line 15: Why do we use the last one? Because we assume that it's the latest deployed backstop pool?

nejcm: Yeah. This is not correct either way. We need a proper way of filtering the pools but I do believe the most recent comes in last so for testing this will be more correct.

ebma (approved):

src/components/Asset/Price/index.tsx
```diff
@@ -4,7 +4,7 @@ import { usePriceFetcher } from '../../../hooks/usePriceFetcher';
 import { Skeleton } from '../../Skeleton';
 
 export type TokenPriceProps = {
-  address: string;
+  address?: string;
   symbol: string;
   //amount?: number;
```
gonzamontiel, line 9: can we remove amount then ?

gonzamontiel (approved):

PR #292, opened 2023-11-07
feat: 288 allow users to withdraw from backstop pool in other assets

nejcm:
On the backstop pool withdraw UI you are now able to select in which you wish to withdraw your liquidity. Currently the picker contains only one asset as the swap pools do not have enough surplus atm.


gonzamontiel:
hey @nejcm is thi ready for review, then ?

src/services/walletConnect/index.ts
```diff
@@ -1,11 +1,11 @@
-import { SignerPayloadJSON } from '@polkadot/types/types';
+import { Signer } from '@polkadot/types/types';
 import { WalletAccount } from '@talismn/connect-wallets';
 import type { SessionTypes } from '@walletconnect/types/dist/types/sign-client/session';
 import UniversalProvider from '@walletconnect/universal-provider';
 import logo from '../../assets/wallet-connect.svg';
 import { config } from '../../config';
 
-// TODO: improve this
+// TODO: improve (missing methods)
```
gonzamontiel, line 8: Is this referred to the methods tagged as 'Unused' ? Do we need them? Is yes, I agree with the TODO, otherwise I would remove it.
```
        getAccounts: () => Promise.resolve([]), // Unused
        subscribeAccounts: () => undefined, // Unused
        transformError: (err: Error) => err, // Unused
```

src/services/walletConnect/index.ts
```diff
@@ -48,7 +48,7 @@ export const walletConnectService = {
       address: accounts[0],
       source: 'walletConnect',
       name: 'WalletConnect',
-      signer,
+      signer: signer as WalletAccount['signer'], // TODO: improve - not type safe
```
gonzamontiel, line 51: How would you improve this ? I personally don't agree with leaving TODO's around the code because, usually, they are never addressed. A better way, if you already know a way to improve this piece, but it is out of scope of this PR, is to create an issue and link it here.

nejcm: The problem currently is that there are some conflicting types between some libraries for the Signer type. This causes conflicts and TS erros when trying to use it. Now this is out of scope for this PR and requires a bit of time investment to fix it. Not sure about creating issues in zenhub for each small improvement in the code as it will cause a lot of noise. But will leave that up to you.

nejcm:
> I left a couple of minor comments. My bigger concern here is that the Nabla code is getting more and more away from our code, in terms of reusable components like Asset selector, dialogs, design in general. Maybe is out of scope of this PR, but at some point we would need to unify this to avoid duplicating efforts when there is an app-wise change on those components.

I would say it is out of scope for this PR but the asset selector used here is the shared one. The only difference is that it was extended to support different types. Regarding the design. I did check figma recently and I did see things have changed for Nabla but I did not manage to reflect those changes yet.

src/services/walletConnect/index.ts
gonzamontiel: imho, all small, medium and large improvements, should have a place to be tracked, and have a description/short explanation. Different thing is if we are going to fix that or not, but at least we know they exist. Once you have a big enough project, those issues might become more relevant and the priority to fix them may also arise.

nejcm:
@gonzamontiel Merge when you wish. I dont have perms.

PR #295, opened 2023-11-14
feat: Update transfer dialogs design

gonzamontiel:
- [x] General dialog changes / refactor
- [x] To Pendulum Completed Dialog (both themes)
- [x] To Pendulum Cancelled Dialog (both themes)
- [x] To Pendulum Pending Dialog (both themes)

### DIFF
--- a/src/pages/bridge/TransferDialog.tsx
+++ b/src/pages/bridge/TransferDialog.tsx
@@ -106,7 +106,7 @@ function BaseTransferDialog(props: BaseTransferDialogProps) {
                 />
               </div>
               <div className="flex flex-row justify-between">
-                <div className="text-sm">Vault PEN Address ({tenantNameCapitalized})</div>
+                <div className="text-sm">Vault Address ({tenantNameCapitalized})</div>
                 <CopyableAddress
                   inline={true}
                   className="text-sm p-0"
@@ -116,7 +116,7 @@ function BaseTransferDialog(props: BaseTransferDialogProps) {
               </div>
               {vaultStellarPublicKey && (
                 <div className="flex flex-row justify-between">
-                  <div className="text-sm">Vault PEN Address</div>
+                  <div className="text-sm">Vault Address (Stellar)</div>
                   <CopyableAddress
                     inline={true}
                     className="text-sm p-0"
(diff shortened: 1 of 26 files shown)


gonzamontiel:
Somehow, after merging staging the styles in the branch are broken. Example: 
Screenshot from 2023-11-23 18 23 18
The overlay doesn't show properly on the dialog. 
I'm trying to figure out the culprit.

src/pages/bridge/TransferDialog.tsx
```diff
@@ -69,66 +76,72 @@ function BaseTransferDialog(props: BaseTransferDialogProps) {
...
                   variant="short"
-                  publicKey={vaultStellarPublicKey}
+                  publicKey={destinationStellarAddress}
                 />
               </div>
-            )}
-            {showMemo && (
               <div className="flex flex-row justify-between">
-                <div className="text-sm">Memo</div>
+                <div className="text-sm">Vault PEN Address ({tenantNameCapitalized})</div>
```
ebma, line 109: We should not call it PEN address. I would actually suggest we completely remove that and only call it `Vault Address ({tenant})`.

src/pages/bridge/TransferDialog.tsx
```diff
@@ -69,66 +76,72 @@ function BaseTransferDialog(props: BaseTransferDialogProps) {
...
                   className="text-sm p-0"
                   variant="short"
-                  publicKey={expectedStellarMemo}
+                  publicKey={transfer.original.vault.accountId.toString()}
                 />
               </div>
-            )}
+              {vaultStellarPublicKey && (
+                <div className="flex flex-row justify-between">
+                  <div className="text-sm">Vault PEN Address</div>
```
ebma, line 119: This is the vault's Stellar address and not a PEN address.

ebma (changes requested):
I know that it's phrased similarly in the designs but we should change the namings.

ebma (approved):
Looks good. But I noticed that something is wrong with the pop up for the active account in the top-right hand corner. It doesn't show properly anymore, not sure if this was introduced in this PR or another

To reproduce, connect to a wallet account and then click on your account in the top right corner. It will show a pop-up but it seems like it is placed above the button and not below it.

PR #301, opened 2023-11-23
fix: lint issues

nejcm:


src/components/Transaction/Settings/index.tsx
```diff
@@ -8,7 +8,7 @@ export interface TransactionSettingsProps {
   deadlineProps?: React.HTMLAttributes<HTMLInputElement>;
 }
 
-const inputCls = 'bg-neutral-100 dark:bg-neutral-900 text-right text-neutral-600 dark:text-neutral-200';
+const inputCls = 'bg-neutral-100 dark:bg-neutral-900 text-right text-neutral-600 dark:text-neutral-200 input-sm';
```
gonzamontiel, line 11: Thanks for the fix, but this is not the actual fix. Changing the type of `deadlineProps` to `InputProps` from daisyui will. I applied this changes to https://github.com/pendulum-chain/portal/pull/295 already :)

gonzamontiel:
I added a comment, thanks for the PR, it works but it doesn't solve the real problem. I will close then.

PR #302, opened 2023-11-24
feat: Add NGN, EUR and AUD assets for Spacewalk

### LINKED ISSUE #290: UI - add new assets (NGN, EUR and AUD) for the Spacewalk bridge
### Context
As there is a spacewalk public launch coming on Amplitude, the new assets should we shown on the portal UI. 
This ticket requests the integration of new assets (NGN, EUR and AUD) into the Spacewalk portal's user interface.

### Describe the feature you'd like:
* Integrate Assets into Spacewalk Portal UI
* Design and implement the user interface (UI) for Spacewalk portal to include NGNC, EURC, and AUDD assets.
* Ensure that users can easily access, view, and interact with these newly added assets.
* This is to be done only on the Amplitude version of spacewalk

### Icons
https://drive.google.com/drive/folders/15g9YrUqFs8PmWyMV65PRFGj4Z97F2N6t?usp=drive_link

gonzamontiel:
- [x] Add assets icons
- [x] Add prices (from DIA oracle)

Closes #290


ebma (commented):
Let's also show these new assets in the portfolio of the dashboard

gonzamontiel:
> Let's also show these new assets in the portfolio of the dashboard

Hmm, they are not hardcoded in the Portfolio. The tokens are fetched from spacewalk, so this PR adds support for the new assets: like, when they are fetched, we have icons and prices to show. Is there something missing that they do not show and they should?

ebma:
Oh, you are right. I double-checked and they don't show because the vaults for these new tokens are not registered yet. And the vaults are used to derive the assets important to us, see [here](https://github.com/pendulum-chain/portal/blob/3084c38ba14872eceb28ffd04fb62e8baca8160a/src/hooks/useBalances.ts#L27-L37). Nevermind then.

ebma (approved):

PR #303, opened 2023-11-24
fix: Update disclaimer text for Spacewalk

### LINKED ISSUE #300: Update the disclaimer text for bridge UI
## Context

We want to update the disclaimer text for Amplitude version of spacewalk bridge

### Text to be updated

- Issue Fee: 0.10% of the transaction amount.
- Redeem Fee: 0.10% of the transaction amount.
- Security deposit: 0.50% of the transaction amount.
- Max issuable amount: 20,000 USD
- Estimated time for issuing: 2mins - 3hrs (only applicable after the user makes the stellar txn to the vault)

gonzamontiel:
Closes #300

### DIFF
--- a/src/pages/bridge/Issue/index.tsx
+++ b/src/pages/bridge/Issue/index.tsx
@@ -10,6 +10,7 @@ import { useNodeInfoState } from '../../../NodeInfoProvider';
 import From from '../../../components/Form/From';
 import OpenWallet from '../../../components/Wallet';
 import { getErrors, getEventBySectionAndMethod } from '../../../helpers/substrate';
+import { useFeePallet } from '../../../hooks/spacewalk/fee';
 import { RichIssueRequest, useIssuePallet } from '../../../hooks/spacewalk/issue';
 import useBridgeSettings from '../../../hooks/spacewalk/useBridgeSettings';
 import { decimalToStellarNative, nativeToDecimal } from '../../../shared/parseNumbers';
@@ -17,13 +18,6 @@ import { FeeBox } from '../FeeBox';
 import { ConfirmationDialog } from './ConfirmationDialog';
 import Disclaimer from './Disclaimer';
 import { getIssueValidationSchema } from './IssueValidationSchema';
-
-const disclaimerText = `• Issue Fee: 0.10% of the transaction amount.
-  • Redeem Fee: 0.10% of the transaction amount.
-  • Security deposit: 0.50% of the transaction amount.
-  • Max issuable amount: 20,000 USD.
-  • Estimated time for issuing: 2 mins to 3 hrs (after submitting the Stellar payment to the vault).
-`;
 
 interface IssueProps {
   network: string;
@@ -47,7 +41,7 @@ function Issue(props: IssueProps): JSX.Element {
   const { walletAccount, dAppName } = useGlobalState();
   const { api } = useNodeInfoState().state;
   const { selectedVault, selectedAsset, setSelectedAsset, wrappedAssets } = useBridgeSettings();
-
+  const { issueFee, redeemFee, issueGriefingCollateral } = useFeePallet().getFees();
   const maxIssuable = nativeToDecimal(selectedVault?.issuableTokens || 0).toNumber();
 
   const { handleSubmit, watch, register, formState, setValue } = useForm<IssueFormValues>({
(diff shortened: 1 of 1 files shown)


src/pages/bridge/Issue/index.tsx
```diff
@@ -5,21 +5,25 @@ import { useCallback, useMemo, useState } from 'preact/hooks';
...
 import { ConfirmationDialog } from './ConfirmationDialog';
 import Disclaimer from './Disclaimer';
 import { getIssueValidationSchema } from './IssueValidationSchema';
 
-const disclaimerText =
-  '1. Sed ut perspiciatis unde omnis iste natus error sit voluptatem\n2. accusantium doloremque laudantium, totam rem aperiam eaque \n3. quae ab illo inventore veritatis et quasi architecto beatae vitae dicta\n4. explicabo. Nemo enim ipsam voluptatem quia voluptas sit aspe';
+const disclaimerText = `• Issue Fee: 0.10% of the transaction amount.
+  • Redeem Fee: 0.10% of the transaction amount.
+  • Security deposit: 0.50% of the transaction amount.
+  • Max issuable amount: 20,000 USD.
```
ebma, line 24: Maybe we should not mention this or rephrase it? It's not wrong but ambiguous and probably confuses more than it helps. Maybe we rephrase it to 'Total issuable amount: tokens worth 20,000 USD'

src/pages/bridge/Issue/index.tsx
```diff
@@ -5,21 +5,25 @@ import { useCallback, useMemo, useState } from 'preact/hooks';
...
+  • Estimated time for issuing: 2mins - 3hrs (applicable after the user makes the Stellar transaction to the vault).
```
ebma, line 25:
```suggestion
  • Estimated time for issuing: 2 mins to 3 hrs (after submitting the Stellar payment to the vault).
```

ebma (commented):
For issue and redeem fee, we might set them to 0.00 quite soon, and then we'd need to update this text again. 

How about for the disclaimer text, we derive these percentages from the actual on-chain values? @gonzamontiel they should already be available from the fee pallet hook. We just need to convert the `issueFee`, `issueGriefingCollateral` and `redeemFee` to percentages by dividing them by 10e18. But we should already have a function for this in our number conversion helpers.

gonzamontiel:
> For issue and redeem fee, we might set them to 0.00 quite soon, and then we'd need to update this text again.
>
> How about for the disclaimer text, we derive these percentages from the actual on-chain values? @gonzamontiel they should already be available from the fee pallet hook. We just need to convert the `issueFee`, `issueGriefingCollateral` and `redeemFee` to percentages by dividing them by 10e18. But we should already have a function for this in our number conversion helpers.

It actually seems the right thing to do, I will implement it and come back to you. I will modify also the total issuable amount text that you mentioned above.

src/pages/bridge/Issue/index.tsx
gonzamontiel: I think the text is accurate but, but I would include the clarification before the `:` , something like:
`Total issuable amount (in USD): 20,000 USD`

src/pages/bridge/Issue/index.tsx
```diff
@@ -58,6 +56,16 @@ function Issue(props: IssueProps): JSX.Element {
     return amount ? decimalToStellarNative(amount) : Big(0);
   }, [amount]);
 
+  const disclaimerText = useMemo(
+    () =>
+      `• Issue Fee: ${issueFee.mul(100)}% of the transaction amount.
+       • Redeem Fee: ${redeemFee.mul(100)}% of the transaction amount.
+       • Security deposit: ${issueGriefingCollateral.mul(100)}% of the transaction amount.
+       • Total issuable amount (in USD): 20,000 USD.
```
ebma, line 64: Nit: If we say it's 'in USD' in the brackets, we don't need to mention it again after the value. 
```suggestion
       • Total issuable amount (in USD): 20,000.
```

ebma (approved):
Looks good overall 👍

src/pages/bridge/Issue/index.tsx
gonzamontiel: I prefer redundancy here.

PR #315, opened 2023-12-13
feat: Metamask Snap integration

### LINKED ISSUE #257: Metamask snap Integration
### Context
So Metamask has built snaps that allows metamask to interact with dApps apart from the ethereum ecosystem. We would like to integrate this into our Portal. So metamask users can interact with our Portal. 

## Requirements

- [ ] Here is the detail description of the integration [process](https://github.com/chainsafe/metamask-snap-polkadot/wiki#integration)
- [ ] Show Metamask as option in the wallet connect dialog box
- [ ] Icon - 
image

gonzamontiel:
This basically applies this [Snap extension](https://github.com/ChainSafe/metamask-snap-polkadot/) for Polkadot.

- [x] Bump some polkadot dependencies
- [x] Add UI for the button/dialog according to our existing talisman dialog
- [x] Create a Wallet Account with the data provided by the snap API
- [x] Test balances and sign transactions

Closes #257


src/services/metamask/metamask.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+}
+
+export function hasMetaMask(): boolean {
+  if (!window.ethereum) {
+    return false;
+  }
+  return window.ethereum.isMetaMask;
+}
+export const defaultSnapId = 'local:http://localhost:8081';
```
ebma, line 26: Do we need this?

ebma (commented):
The installation and everything worked well. 
I have two accounts imported in my metamask though and I wonder why only one account is selectable in the selector. Is it the same for you @prayagd @gonzamontiel? 
Besides that, sometimes the request for signing the transaction does not go through. I can't reproduce this all the time but I tested this with staking on Foucoco (first had to top up the account that is shown by metamask of course) and sometimes the request for signing times out when you click on the 'confirm' button to finalize the delegation. If you click the button again after the request timed out it generally works. Any clue why it might fail sometimes @gonzamontiel?
It would also be nice if the account automatically reconnects when you refresh the page like we do with the other wallet extensions.

gonzamontiel:
hey @ebma, I go inline.

> The installation and everything worked well. I have two accounts imported in my metamask though and I wonder why only one account is selectable in the selector. Is it the same for you @prayagd @gonzamontiel?

The thing is that this integration works in a very weird way, and the docs are not very clear about this. But the extension will _create_ a Polkadot account for you associated to your Metamask extension. Because snaps are a way to extend Metamask functionality _beyond_ the wallet itself. Maybe it changed, but afaik, you can only connect to the account it was created by the extension. 

> Besides that, sometimes the request for signing the transaction does not go through. I can't reproduce this all the time but I tested this with staking on Foucoco (first had to top up the account that is shown by metamask of course) and sometimes the request for signing times out when you click on the 'confirm' button to finalize the delegation. If you click the button again after the request timed out it generally works. Any clue why it might fail sometimes @gonzamontiel?

Sometimes I noticed the same with low internet connection, I think it is a request that takes too long, and we might need to push that limit. What I guess happens is that the extension takes even longer to sign than a normal account, and then this timeout happens more often. But I gues we need to revisit the Staking process and it should be done in a separete PR.

> It would also be nice if the account automatically reconnects when you refresh the page like we do with the other wallet extensions.

I will look into this.

src/services/metamask/metamask.ts
gonzamontiel: Removed.

gonzamontiel:
> > It would also be nice if the account automatically reconnects when you refresh the page like we do with the other wallet extensions.
> 
> I will look into this.

This is addressed, please review again @ebma

ebma:
The account now automatically reconnects, or at least it's shown in the wallet selector in the top-right hand corner. But when you try to submit something after refreshing the page, it will fail with the following error message. 
image

Sharqiewicz:
When I choose to log in with metamask and I see the modal with the Metamask address, I can no longer go back to select another wallet. What do you think about adding a button so it is possible to undo the wallet selection?
Screenshot 2024-02-20 at 13 04 57

src/services/metamask/metamask.ts
```diff
@@ -0,0 +1,140 @@
...
+      return 'kusama';
+    default:
+      return 'kusama';
+  }
+}
+
+export async function installPolkadotSnap(relayChain: SnapNetworks): Promise<boolean> {
+  try {
+    await enablePolkadotSnap({ networkName: relayChain });
+    console.info('Snap installed!!');
```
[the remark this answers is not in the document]
gonzamontiel: In this case it is intended, we try not to flood the code, but some feedback about that the snaps or some providers are connected seems harmless, ain't it ?

gonzamontiel:
> When I choose to log in with metamask and I see the modal with the Metamask address, I can no longer go back to select another wallet. What do you think about adding a button so it is possible to undo the wallet selection? Screenshot 2024-02-20 at 13 04 57


Done, now it looks like this:

Screenshot from 2024-02-20 17 27 59

Sharqiewicz (approved):
LGTM 🚀

PR #322, opened 2023-12-20
fix: Fix Back to Stellar confirmation dialog

### LINKED ISSUE #320: Back to stellar request text greyed out on pop-up
### Describe the bug

When a user creates a to stellar request, the pop-up shows greyed out text which does not fit the design criteria

### Reproduction

1. Go to /spacewalk/bridge
2. Click on to stellar, fill out the details and create a request
3. A pop-up will appear, here half of the text is greyed out. More context in the screenshot

### Expected behavior

All the text on the pop-up should be of the same color

Adding some more notes, to be fixed in the same ticket
1. Remove the the field stellar destination address
2. Change the text above button to "We will update the transaction when PEN payment is executed. This typically takes only a few minutes but may sometimes take up to 6 hours"
3. Clicking on "View Progress" should re-direct to `/spacewalk/transfers` page

### Screenshots
image

gonzamontiel:
Closes #320

### DIFF
--- a/src/pages/bridge/Redeem/ConfirmationDialog.tsx
+++ b/src/pages/bridge/Redeem/ConfirmationDialog.tsx
@@ -1,8 +1,8 @@
-import { useMemo } from 'preact/compat';
 import { Button, Modal } from 'react-daisyui';
-import { CopyableAddress, PublicKey } from '../../../components/PublicKey';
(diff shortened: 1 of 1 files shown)


prayagd (approved):
Perfecto, good to go

PR #326, opened 2024-01-04
Add Nabla UI

nejcm:
This PR swaps the polkadot-api contract calls with pendulum-solang library.


TorstenStueber:
@nejcm I see that you are still adding commits to this PR. Is this already ready to be reviewed or still work in progress?

nejcm:
I had to fix the failing testa and refactor some folders. This is ready to merge. I dont have perms to do so though.

TorstenStueber:
Okay, I will review this ticket and test locally. The PR is quite massive, so this will take a little time.

nejcm:
So it looks like a lot mostly because of the graphql generated files being changed. If you need and info or help from my side I am available.

src/SharedProvider.tsx
```diff
@@ -7,6 +7,8 @@ const SharedProvider = ({ children }: { children: ComponentChildren }) => {
   const { api } = useNodeInfoState().state;
   const { signer, address } = useGlobalState().walletAccount || {};
 
+  console.log('Provider', api, signer, address);
```
Sharqiewicz, line 10: nit: Do we need this console.log?

src/components/Layout/NavCollapseButtonContent.tsx
```diff
@@ -8,20 +8,22 @@ interface NavButtonContentProps {
...
-        <Lottie options={item.title.lottieOptions} isStopped={!isPlaying} {...item.title.componentOptions} />
-      </span>
-    ) : (
-      <span>{item.title}</span>
-    )}
-    {item.suffix}
-  </>
-);
+export const NavCollapseButtonContent: React.FC<NavButtonContentProps> = ({ item, isPlaying }) => {
+  return (
```
Sharqiewicz, line 12: nit: We can just return the component like 
```
=> (
<>...
</>
);
```
Instead of using return keyword

src/components/nabla/Pools/Backstop/AddLiquidity/useAddLiquidity.ts
```diff
@@ -1,56 +1,102 @@
...
-      queryClient.refetchQueries([cacheKeys.backstopPools, indexerUrl]);
+    mutateOptions: {
+      onError: () => {
+        // ? log error - alert not needed as the transaction modal displays the error
+      },
+      onSuccess: () => {
+        form.reset();
+        balanceQuery.refetch();
+        depositQuery.refetch();
+        setTimeout(() => {
```
Sharqiewicz, line 62: Using `setTimeout` to refetch queries is not the most efficient way to handle data refetching. 
Instead we could maybe make the onSuccess `async` and put `await` before balanceQuery and depositQuery to make sure they're finished before the `queryClient.refetchQueries` is executed.
We could also try to put the `queryClient.refetchQueries([cacheKeys.backstopPools, indexerUrl])` in `onSettled` as it runs after `onSuccess` ( but it is also triggered after `onError` )

src/components/nabla/Pools/Backstop/WithdrawLiquidity/useBackstopDrain.ts
```diff
@@ -0,0 +1,65 @@
...
+import { getValidSlippage } from '../../../../../helpers/transaction';
+import { decimalToRaw } from '../../../../../shared/parseNumbers/metric';
+import { NablaInstanceBackstopPool, NablaInstanceSwapPool } from '../../../../../hooks/nabla/useNablaInstance';
+import { useContractWrite } from '../../../../../hooks/nabla/useContractWrite';
+import { WithdrawLiquidityValues } from './useWithdrawLiquidity';
+import { UseQuoteResult } from '../../../../../hooks/nabla/useQuote';
+
+export type UseBackstopDrainProps = {
+  backstopPool: NablaInstanceBackstopPool;
+  selectedSwapPool: NablaInstanceSwapPool | undefined;
```
Sharqiewicz, line 14: nit: `selectedSwapPool?:NablaInstanceSwapPool`

src/main.tsx
```diff
@@ -14,11 +13,11 @@ import SharedProvider from './SharedProvider';
 
 const queryClient = new QueryClient();
 
-const localStoragePersister = createSyncStoragePersister({ storage: window.localStorage });
+/* const localStoragePersister = createSyncStoragePersister({ storage: window.localStorage });
 persistQueryClient({
   queryClient,
   persister: localStoragePersister,
-});
+}); */
```
ebma, line 20: Why are we removing this?

src/components/nabla/Pools/Backstop/WithdrawLiquidity/useWithdrawLiquidity.ts
```diff
@@ -2,91 +2,159 @@ import { yupResolver } from '@hookform/resolvers/yup';
...
-          id: '',
-          token: {
-            ...token,
-            id: '',
-          },
-        } as SwapPool,
-      ].concat(swapPools || []),
-    [swapPools, token],
+  const selectedPool = useMemo<NablaInstanceSwapPool | undefined>(
+    () => swapPools.find((t) => t.id === address)!,
```
Sharqiewicz, line 92: Is the "!" at the end necessary?

Sharqiewicz: It's generally best to avoid using TypeScript's non-null assertion as it can lead to runtime errors if our assumptions are incorrect. What do you think about adding a check for undefined here?

src/components/nabla/Pools/Backstop/WithdrawLiquidity/useWithdrawLiquidity.ts
```diff
@@ -2,91 +2,159 @@ import { yupResolver } from '@hookform/resolvers/yup';
...
+  const { refetch: balanceRefetch } = balanceQuery;
+  const { refetch: depositRefetch } = depositQuery;
+  const { refetchQueries } = queryClient;
   const onWithdrawSuccess = useCallback(() => {
     reset();
     balanceRefetch();
     depositRefetch();
-    queryClient.refetchQueries([cacheKeys.backstopPools, indexerUrl]);
-  }, [balanceRefetch, depositRefetch, indexerUrl, queryClient, reset]);
+    setTimeout(() => {
```
Sharqiewicz, line 122: Could we apply the same fix about not using setTimeout here as well? It would help us maintain consistency and avoid potential issues across the application.

src/components/nabla/Pools/Backstop/WithdrawLiquidity/useWithdrawLiquidity.ts
```diff
@@ -103,19 +171,17 @@ export const useWithdrawLiquidity = (pool: BackstopPool) => {
...
-    form,
-    isSwapPoolWithdraw,
-    pools,
     selectedPool,
-    swapPoolWithdraw,
-    tokenModal,
+    tokenToReceive,
+    withdrawalQuote,
     toggle,
+    onSubmit: form.handleSubmit(selectedPool !== undefined ? backstopDrain.onSubmit : backstopWithdraw.onSubmit),
```
Sharqiewicz, line 184: I wanted to confirm whether we are only focusing on `undefined` value and not considering `null` value in this check. Is that correct?

src/hooks/nabla/useTokenOutAmount.ts
```diff
@@ -0,0 +1,132 @@
...
+      parseError: (error) => {
+        switch (error.type) {
+          case 'error':
+            return 'Something went wrong';
+          case 'panic':
+            return error.errorCode === 0x11 ? 'The input amount is too large' : 'Something went wrong';
+          case 'reverted':
+            return error.description === 'SwapPool: EXCEEDS_MAX_COVERAGE_RATIO'
+              ? 'The input amount is too large'
+              : 'Something went wrong';
```
ebma, line 115: Can we use an error message other than `'The input amount is too large'`? As a user, I don't know how it is too large, and I don't even know which amount this is referring to. It's not obvious that the error is about the coverage ratio of the pool.

ebma (commented):
Generally, I think the code is okay. It's not great but mostly because the implementation of the Nabla UI happened in parallel but in isolation so we now have a confusing structure and some duplicated components that we should streamline. 

However, I think we should rather address the problem of duplicate components and bad structure of our code files in a separate PR with the sole purpose of fixing those things. Then we can also discuss and agree how this perfect structure should look like.

src/components/nabla/Pools/Swap/AddLiquidity/useAddLiquidity.ts
```diff
@@ -1,60 +1,102 @@
...
-      queryClient.refetchQueries([cacheKeys.swapPools, indexerUrl]);
+    mutateOptions: {
+      onError: () => {
+        // ? log error - alert not needed as the transaction modal displays the error
+      },
+      onSuccess: () => {
+        form.reset();
+        balanceQuery.refetch();
+        depositQuery.refetch();
+        setTimeout(() => {
```
Sharqiewicz, line 62: Could we apply the same fix about not using setTimeout here as well? It would help us maintain consistency and avoid potential issues across the application.

src/components/nabla/Pools/Swap/WithdrawLiquidity/useWithdrawLiquidity.ts
```diff
@@ -1,70 +1,129 @@
...
-      queryClient.refetchQueries([cacheKeys.swapPools, indexerUrl]);
+    mutateOptions: {
+      onError: () => {
+        // ? log error - alert not needed as the transaction modal displays the error
+      },
+      onSuccess: () => {
+        form.reset();
+        balanceQuery.refetch();
+        depositQuery.refetch();
+        setTimeout(() => {
```
Sharqiewicz, line 77: Could we apply the same fix about not using setTimeout here as well? It would help us maintain consistency and avoid potential issues across the application.

src/components/nabla/common/NumberInput.tsx
```diff
@@ -0,0 +1,34 @@
+import { HTMLAttributes, useCallback } from 'react';
```
Sharqiewicz, line 1: We shoudn't import from react, we should import from 'preact/compat' instead

src/components/nabla/Pools/Backstop/WithdrawLiquidity/useBackstopDrain.ts
```diff
@@ -0,0 +1,65 @@
+import { useCallback } from 'react';
```
Sharqiewicz, line 1: We shouldn't import from `react`, we should import from `preact/compat` instead

src/hooks/nabla/useQuote.ts
```diff
@@ -0,0 +1,107 @@
+import { Big } from 'big.js';
+
+import { activeOptions } from '../../constants/cache';
+import { multiplyByPowerOfTen } from '../../shared/parseNumbers/metric';
+import { MessageCallErrorResult, useContractRead } from './useContractRead';
+import { useDebouncedValue } from '../useDebouncedValue';
+import { FieldValues, UseFormReturn } from 'react-hook-form';
+import { useEffect, useMemo } from 'preact/hooks';
+import { ContractBalance, parseContractBalanceResponse } from '../../helpers/contracts';
+import { useCallback } from 'react';
```
Sharqiewicz, line 10: We shouldn't import from `react`, we should import from `preact/compat` instead

src/components/nabla/Pools/Backstop/WithdrawLiquidity/useBackstopWithdraw.ts
```diff
@@ -1,40 +1,51 @@
 import { useCallback } from 'react';
```
Sharqiewicz, line 1: We shouldn't import from `react`, we should import from `preact/compat` instead

src/components/nabla/Swap/ApprovalSubmit.tsx
```diff
@@ -0,0 +1,35 @@
+import { Button } from 'react-daisyui';
+import Big from 'big.js';
+
+import { useGetAppDataByTenant } from '../../../hooks/useGetAppDataByTenant';
+import { NablaInstanceToken } from '../../../hooks/nabla/useNablaInstance';
+import { TokenApproval } from '../common/TokenApproval';
+
+export interface ApprovalProps {
+  token: NablaInstanceToken | undefined;
```
Sharqiewicz, line 9: nit: `token?: NablaInstanceToken`

src/components/nabla/Swap/ApprovalSubmit.tsx
```diff
@@ -0,0 +1,35 @@
...
+  disabled: boolean;
+  fromAmount: Big | undefined;
```
Sharqiewicz, line 11: nit: `fromAmount?: Big`

src/components/nabla/Swap/From.tsx
```diff
@@ -0,0 +1,99 @@
+import { ChevronDownIcon } from '@heroicons/react/20/solid';
+import { Fragment } from 'preact';
```
Sharqiewicz, line 2: Instead of importing and using Fragment, we might consider using the shorthand `<> </>` syntax for a cleaner approach. What do you all think?

src/components/nabla/Swap/From.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { NablaInstanceToken } from '../../../hooks/nabla/useNablaInstance';
+import { NablaTokenPrice } from '../common/NablaTokenPrice';
+import { fractionOfValue } from '../../../shared/parseNumbers/metric';
+import { AmountSelector } from '../common/AmountSelector';
+import { UseContractReadResult } from '../../../hooks/nabla/useContractRead';
+import { ContractBalance } from '../../../helpers/contracts';
+import { TokenBalance } from '../common/TokenBalance';
+
+interface FromProps<FormFieldValues extends FieldValues, TFieldName extends FieldPath<FormFieldValues>> {
+  fromToken: NablaInstanceToken | undefined;
```
Sharqiewicz, line 17: nit: fromToken?: NablaInstanceToken

src/components/nabla/Swap/From.tsx
```diff
@@ -0,0 +1,99 @@
...
+        <div className="flex gap-1 text-sm">
+          {fromTokenBalance !== undefined && (
+            <Fragment>
+              <span className="mr-1">
+                Your Balance: <TokenBalance query={fromTokenBalance} symbol={fromToken?.symbol}></TokenBalance>
+              </span>
+              <button
+                className="text-primary hover:underline"
+                onClick={() => {
+                  if (fromTokenBalance.data !== undefined) {
```
Sharqiewicz, line 74: I wanted to confirm whether we are only focusing on `undefined` value and not considering `null` value in this check. Is that correct?

src/components/nabla/Swap/From.tsx
```diff
@@ -0,0 +1,99 @@
...
+                  }
+                }}
+                type="button"
+              >
+                50%
+              </button>
+              <button
+                className="text-primary hover:underline"
+                onClick={() => {
+                  if (fromTokenBalance.data !== undefined) {
```
Sharqiewicz, line 85: I wanted to confirm whether we are only focusing on `undefined` value and not considering `null` value in this check. Is that correct?

src/components/nabla/common/TransactionSettingsDropdown.tsx
```diff
@@ -0,0 +1,86 @@
...
+            />
+            <span className="dark:text-neutral-400 text-neutral-600">minutes</span>
+          </div>
+        </div>
+      )}
+    </div>
+  );
+};
+
+export function TransactionSettingsDropdown(props: TransactionSettingsProps & { button?: ReactNode }) {
```
Sharqiewicz, line 69: Perhaps we could export this to another file for better modularity and easier maintenance. What do you think?

src/components/nabla/common/TransactionSettingsDropdown.tsx
```diff
@@ -0,0 +1,86 @@
+import { Cog8ToothIcon } from '@heroicons/react/24/outline';
+import { Button, Dropdown, Input } from 'react-daisyui';
```
Sharqiewicz, line 2: nit: Can we just import the `InputProps` from `react-daisyui` directly?

src/components/nabla/common/TokenApproval.tsx
```diff
@@ -0,0 +1,62 @@
...
+  });
+
+  const onClick = useCallback(() => {
+    if (state === ApprovalState.PENDING || decimalAmount === undefined) return;
+
+    const nativeAmount = multiplyByPowerOfTen(decimalAmount, decimals);
+    mutate([spender, nativeAmount.round(0, 0).toFixed()]);
+  }, [state, mutate, decimalAmount, decimals, spender]);
+
+  if (state === ApprovalState.APPROVED || !enabled) return <>{children}</>;
```
Sharqiewicz, line 44: Perhaps we can just `return children` instead of `return <>{children}</>`. This could simplify our component slightly. What do you think?

src/components/nabla/common/NumberInput.tsx
Sharqiewicz, line 1: We have already implemented a dedicated input for handling numbers in our project. Previously, we faced several issues with inputMode="numeric", and after extensive discussions and recommendations, we decided to create a specialized field. You can find this implementation in `/src/components/Form/From/InputField/index.tsx`. Please take a look if it can be used instead of NumberInput

src/components/nabla/common/NablaTokenPrice.tsx
```diff
@@ -0,0 +1,21 @@
...
+  address: string;
+  prefix?: ReactNode;
+  fallback?: ReactNode;
+};
+
+export function NablaTokenPrice({ address, prefix = null, fallback = null }: TokenPriceProps): JSX.Element | null {
+  const { data, isLoading } = useNablaTokenPrice(address);
+  if (isLoading) return <NumberLoader />;
+
+  if (data === undefined) return <>{fallback}</>;
```
Sharqiewicz, line 14: Perhaps we can just `return fallback` instead of `return <>{fallback}</>`

src/components/nabla/Swap/index.tsx
```diff
@@ -19,28 +18,27 @@ const Swap = (props: UseSwapComponentProps): JSX.Element | null => {
...
-  const progressUi = useMemo(() => {
-    if (swapMutation?.isIdle) return null;
-    const { from: fromV, to: toV, fromAmount = 0, toAmount = 0 } = getValues();
-    const fromAsset = tokensMap?.[fromV];
-    const toAsset = tokensMap?.[toV];
-    return (
-      <p className="text-center text-[--text]">{`Swapping ${fromAmount} ${fromAsset?.symbol} for ${toAmount} ${toAsset?.symbol}`}</p>
-    );
-  }, [getValues, swapMutation?.isIdle, tokensMap]);
+  const submitEnabled = !toAmountQuote.isLoading && toAmountQuote.enabled && Object.keys(errors).length === 0;
```
Sharqiewicz, line 40: I think instead of using `Object.keys(errors).length === 0`, we can use isValid from `formState: { errors, isValid }`

Sharqiewicz (commented):
PR looks good 😎 🚀  , a few small changes to review if they should be applied. I think we are also repeating the code of <input> element in the application, and we should someday create a single source of truth for using inputs in our application, because otherwise we may be repeating code unnecessarily and sometimes this can lead to errors when we miss some validation in one input

src/SharedProvider.tsx
TorstenStueber: No, I was not aware that it was added in this PR. Removed.

src/components/nabla/Pools/Backstop/AddLiquidity/useAddLiquidity.ts
TorstenStueber: This was not put there by me originally. There was a comment somewhere but I removed it. The reason to delay this is not to wait for `balanceQuery` and `depositQuery` to finish. The call to the refetch the `nablaInstance` is totally independent. The first two query directly from the smart contract on the blockchain. The `nablaInstance` comes from the indexer. We observed during tests that the indexer is always a little delayed. For that reason we wait a little to have a higher chance that it already delivers the new data.

src/components/nabla/Pools/Backstop/BackstopPoolModals.tsx
```diff
@@ -1,18 +1,28 @@
...
-import { LiquidityModalProps } from './types';
+
+import { NablaInstanceBackstopPool } from '../../../../hooks/nabla/useNablaInstance';
+import { useModal } from '../../../../services/modal';
+import ModalCloseButton from '../../../Button/ModalClose';
+import AddLiquidity from './AddLiquidity';
+import WithdrawLiquidity from './WithdrawLiquidity';
+
+export const ModalTypes = {
+  AddLiquidity: 2,
```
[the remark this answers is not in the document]
TorstenStueber: Agree. I took this code over but did not bother to improve. I replaced it with a TypeScript solution now and direct string constants.

src/components/nabla/Pools/Backstop/WithdrawLiquidity/useBackstopDrain.ts
TorstenStueber: Both types are almost the same, but you see that I do this style a lot if the value for that parameter is always provided at every call site (instead of being left out sometimes) – I can change if it clashes with the style in the rest of the codebase.

src/components/nabla/Pools/Backstop/WithdrawLiquidity/useWithdrawLiquidity.ts
TorstenStueber: This indeed does not make sense here. Removed.

src/components/nabla/Pools/Backstop/WithdrawLiquidity/useWithdrawLiquidity.ts
TorstenStueber: Here we query the smart contract and don't need to delay as the smart contract will already have its updated value at this point (unlike the indexer).

src/components/nabla/Pools/Backstop/WithdrawLiquidity/useWithdrawLiquidity.ts
TorstenStueber: `selectedPool` can never be `null`, only `undefined`, also ensured by TypeScript.

TorstenStueber:
> Generally, I think the code is okay. It's not great but mostly because the implementation of the Nabla UI happened in parallel but in isolation so we now have a confusing structure and some duplicated components that we should streamline. 
> 
> 
> 
> However, I think we should rather address the problem of duplicate components and bad structure of our code files in a separate PR with the sole purpose of fixing those things. Then we can also discuss and agree how this perfect structure should look like. 
> 
> 
> 
> 

Fully agree!

src/components/nabla/Pools/Swap/AddLiquidity/useAddLiquidity.ts
TorstenStueber: Same comment here, this is the indexer again.

src/components/nabla/Pools/Swap/Redeem/index.tsx
```diff
@@ -1,145 +1,159 @@
...
-    form: { register, setValue },
-  } = useRedeem(data);
-  const deposit = depositQuery.balance || 0;
+    register,
+    setValue,
+    formState: { errors },
+  } = form;
+
+  const totalSupplyOfLpTokens = rawToDecimal(data.totalSupply, data.token.decimals);
+  const submitEnabled = !withdrawalQuote.isLoading && withdrawalQuote.enabled && Object.keys(errors).length === 0;
```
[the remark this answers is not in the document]
TorstenStueber: Are you sure that this works? I figured this is only for validation errors, but here I check for errors I explicitly attach to input fields of the form. I didn't test and the docs weren't clear.

src/components/nabla/Pools/Swap/SwapPoolModals.tsx
```diff
@@ -1,11 +1,22 @@
...
+import { useModal } from '../../../../services/modal';
+import ModalCloseButton from '../../../Button/ModalClose';
+import AddLiquidity from './AddLiquidity';
+import Redeem from './Redeem';
+import WithdrawLiquidity from './WithdrawLiquidity';
+
+import { SwapPoolColumn } from './columns';
+
+export const ModalTypes = {
+  AddLiquidity: 2,
```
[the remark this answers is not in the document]
TorstenStueber: Done (see my other comment)

src/components/nabla/Pools/Backstop/AddLiquidity/useAddLiquidity.ts
Sharqiewicz: Perhaps we could extract this setTimeout for 2000ms to a separate function and reuse it wherever needed. This way, if we need to make changes in the future, we can do so in one place rather than having to search throughout the application. What do you think about this approach?

TorstenStueber: I extracted it to a function `refetchDelayed`.

src/main.tsx
TorstenStueber: This happened in [this commit](https://github.com/pendulum-chain/portal/pull/326/commits/ad4ff0cec093abf0d33ff8c796ea33ab70131c19) and I don't know the reasoning for it either. Do we need this storage persister? Seems like it stores responses to fetch requests (indexer or RPC) in the local storage.

src/components/nabla/Pools/Swap/Redeem/index.tsx
```diff
@@ -1,145 +1,159 @@
...
   const hideCss = !mutation.isIdle ? 'hidden' : '';
   return (
     <div className="text-[initial] dark:text-neutral-200">
       <TransactionProgress mutation={mutation} onClose={mutation.reset}>
-        <PoolProgress symbol={data.token.symbol} amount={amount} />
+        <PoolProgress symbol={data.symbol} amount={amountString} />
       </TransactionProgress>
       <div className={hideCss}>
         <div className="flex items-center gap-2 mt-2 mb-8">
           <Button size="sm" color="ghost" className="px-2" type="button" onClick={() => toggle(undefined)}>
```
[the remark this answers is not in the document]
TorstenStueber: No. Removed.

src/components/nabla/Pools/Swap/WithdrawLiquidity/index.tsx
```diff
@@ -1,38 +1,41 @@
...
-    form: { register, setValue },
-  } = useWithdrawLiquidity(data.id, data.token.id);
-  const deposit = depositQuery.balance || 0;
+    formState: { errors },
+  } = form;
+
+  const totalSupplyOfLpTokens = rawToDecimal(data.totalSupply, data.token.decimals);
+  const backstopBurnIsPossible = BigInt(data.reserve) < BigInt(data.totalLiabilities);
+
+  const submitEnabled = !withdrawalQuote.isLoading && withdrawalQuote.enabled && Object.keys(errors).length === 0;
```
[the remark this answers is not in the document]
TorstenStueber: See my other comment about this.

src/components/nabla/Swap/ApprovalSubmit.tsx
TorstenStueber: See my other comment about this.

src/components/nabla/Swap/From.tsx
TorstenStueber: See my other comment about this.

src/components/nabla/common/TransactionSettingsDropdown.tsx
TorstenStueber: I find it acceptable to leave it in the same file as these two are quite closely coupled.

src/components/nabla/common/NumberInput.tsx
TorstenStueber: I wasn't aware. What is the problem with inputMode="numeric", can you link to that discussion? Might make sense to unify these components but I would not do in this PR anymore.

src/components/nabla/Swap/index.tsx
TorstenStueber: See my other comment.

TorstenStueber:
I addresses add issues in the PR and bugs found during QA. I also added proper token logos and there are no merge conflicts left.

ebma (approved):
Nice, thanks for putting in the effort to fix all the remaining issues @TorstenStueber 👍 
Let's merge this and create a follow-up ticket for refactoring.

src/components/nabla/Pools/Backstop/AddLiquidity/useAddLiquidity.ts
Sharqiewicz: Thank you @TorstenStueber

Sharqiewicz (approved):
Perfect ✅

prayagd:
@TorstenStueber is backstop pool fixed withdrawal fixed?

TorstenStueber:
> @TorstenStueber is backstop pool withdrawal fixed?

@prayagd Yes.

PR #332, opened 2024-01-22
Use different plugin for polyfilling / Fix walletconnect

### LINKED ISSUE #327: Wallet Connect not working
### Describe the bug

Wallet connect not working, the QR code does not show up

### Reproduction

1. Go to portal.pendulumchain.org
2. Click on Connect wallet
3. Click on Wallet connect option
4. It shows Loading text and nothing happens

### Expected behavior

Once clicking on wallet connect, user should be able to see the QR for connecting the wallet using the Mobile app

### Screenshots
Screenshot 2024-01-10 at 12 09 39 PM

image

### Desktop (please complete the following information):
- OS:  MacOS
- Browser: Brave, Chrome, Firefox

ebma:
### What:

The issue about the WalletConnect connection not working is fixed. 

### How:

Apparently the polyfills are causing some issues, see [this](https://github.com/WalletConnect/walletconnect-monorepo/issues/4064#issuecomment-1885607272
) comment. The issue was addressed by changing the polyfills.

### Comments:

It happened to me that a `no matching key. keychain ....` error was shown in the console when trying to sign a transaction. It's not 100% clear to me why this happened but I found some related issue tickets which suggested either updating the dependencies or cleaning local storage. Deleting the walletconnect-related items from local storage and then disconnecting, reconnecting again fixed this issue for me.

Closes #327.


TorstenStueber:
@ebma why do we need to polyfill Node.JS in a frontend application in the first place? This sounds strange to me.

Did you check whether removing this polyfill altogether works as well (see the second paragraph of [this comment](https://github.com/WalletConnect/walletconnect-monorepo/issues/4064#issuecomment-1878549480))?

ebma:
@TorstenStueber good question. I looked it up again and I found [this](https://github.com/pendulum-chain/portal/commit/335fdca5f35be8bc98bc30449b7e4a21ae46546c) commit. Seems like I added the polyfills for the stellar-sdk. Then I checked if there are still errors and indeed, without the polyfills the stellar-sdk throws some errors. 
image

We could try to reduce the polyfills to the bare minimum instead of using the default configuration but I don't think that it's that important. What do you think @TorstenStueber?

prayagd:
@ebma i am able to connect on Amplitude but it fails on Pendulum as the wallet does not load 

I also see and error saying error while connecting the node

TorstenStueber:
@ebma about the polyfills: I remember now and never liked this part of the stellar-sdk, quite bloated and with unnecessary dependencies. Okay, so then we need to keep this.

gonzamontiel (approved):
Fine for me, didn't test.

gonzamontiel:
> I think that's unrelated and more because our RPC nodes on Pendulum are still very slow and reject connections a lot.

I can confirm this, we experience random failures while accessing Pendulum RPC nodes from the portal.

prayagd:
> I think that's unrelated and more because our RPC nodes on Pendulum are still very slow and reject connections a lot.
@ebma can we do anything about this, personally i think its not so nice experience to get that error often

ebma:
Do you still experience issues @prayagd? I tested it just now, refreshing https://portal.pendulumchain.org/pendulum/dashboard quite often but it doesn't fail for me anymore.

But anyways, this is unrelated to this PR so I'll merge it now.

Issue #163, opened 2023-05-24
Redeem - Autofill with personal data


LeoFraq:
### Describe the bug

Filling in the Redeem form tries to autofill with personal credentials

### Reproduction

1. Go to Bridge
2. Click on Back To Stellar
3. Enter stellar address, having saved personal data may be prompted for input

### Expected behavior

The form does not autofill with this data

### Screenshots

![image.png](https://images.zenhubusercontent.com/640f1eb404b2a44433611a9f/22497552-3043-4414-ac92-66429e7c50a3)

gonzamontiel:
Hi @LeoFraq , thanks for the report! very useful to make the portal functions better. I changed the type to 'enhancement' because autofill isn't necessarily expected, and wasn't defined before. But I think we can implement it with low effort.

PR #339, opened 2024-02-16
fix: Turn off autoComplete for Back to Stellar address input

Sharqiewicz:
### What:

Filling in the Redeem form tries to autofill with personal credentials. Expected behaviour is that the form does not autofill with this data.

### How:

Implement autoComplete="off" for the input.

Closes: #163


gonzamontiel:
Okay, I am not sure about this one. What if you regularly paste your stellar address want that autofilled? Anyway, I don't experience this in my browser but if it's very annoying to you, I won't complain.

gonzamontiel (approved):

Issue #296, opened 2023-11-17
Reserve fees when user does max staking


prayagd:
### Context
When user decides to stakes with "MAX" button, the entire PEN balance is considered in that transaction leaving no tokens for the user to execute the transaction i.e no gas fees to execute the transaction

## Acceptance criteria
* On /staking
* User should be able to stake
* On the pop-up when user clicks on "Max"
* Fill in the input field with so that user is left with 0.5 PEN as balance
* For ex. If user has 20 PEN as decides to do "Max" stake, then the input field should automatically fill in with 19.50 and user should be left with 0.50 PEN

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/portal/296?planning-poker) with Zenhub @ebma @gonzamontiel

PR #343, opened 2024-02-19
fix: add balance left for tx approval: staking

Sharqiewicz:
### What:

When user decides to stakes with "MAX" button, the entire PEN balance is considered in that transaction leaving no tokens for the user to execute the transaction i.e no gas fees to execute the transaction

### How:

Add BALANCE_LEFT_FOR_TX_APPROVAL constant variable.

Closes: #296

### DIFF
--- a/src/components/Form/Amount/index.tsx
+++ b/src/components/Form/Amount/index.tsx
@@ -13,6 +13,12 @@ export interface AmountProps {
 }
 
 const BALANCE_LEFT_FOR_TX_APPROVAL = 0.5;
+
+function calculateMaxAmount(max: number) {
+  const maxAmount = roundNumber(max - BALANCE_LEFT_FOR_TX_APPROVAL);
+  return Math.max(0, maxAmount);
+}
+
 const Amount = ({ className, register, max, setValue, error }: AmountProps): JSX.Element | null => {
   return (
     <>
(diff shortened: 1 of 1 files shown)


src/components/Form/Amount/index.tsx
```diff
@@ -45,7 +46,7 @@ const Amount = ({ className, register, max, setValue, assetSuffix, error }: Amou
             </button>
             <button
               className="text-accent-content underline hover:opacity-70 mx-1 font-semibold"
-              onClick={() => setValue(roundNumber(Number(max)))}
+              onClick={() => setValue(roundNumber(Number(max) - BALANCE_LEFT_FOR_TX_APPROVAL))}
```
ebma, line 49: A user that has `0.00` tokens available will now enter `-0.5` when clicking on the 'MAX' button. Let's maybe wrap this with
```suggestion
              onClick={() => setValue(Math.max(roundNumber(Number(max) - BALANCE_LEFT_FOR_TX_APPROVAL, 0))}
```

ebma (changes requested):

src/components/Form/Amount/index.tsx
Sharqiewicz: @ebma updated ✅

ebma (approved):
LGTM 👍

Issue #345, opened 2024-02-19
Features on staging not working


prayagd:
### Describe the bug

Following features which are merged but dont work on the staging branch
- Cannot click on the spacewalk transfer line item
- Only able to see XLM as the asset on the spacewalk bridge dropdown

### Reproduction

Steps to reproduce the behavior: Cannot click on the spacewalk transfer line item

1. Go to '/spacewalk/transfers`
2. Click on any line item from the transfer list
3. Does not open, also asset names not visible in the asset column

Steps to reproduce the behavior: Only able to see XLM as the asset on the spacewalk bridge dropdown

1. Go to '/spacewalk/bridge`
2. on issue tab, open the dropdown
3. See only the XLM token in the dropdown

### Expected behavior

1. Cannot click on the spacewalk transfer line item - Expected is to click on it and it should open up #311 
2. Only able to see XLM as the asset on the spacewalk bridge dropdown - Expected to show all the 7 assets on Spacewalk Amplitude USDC, XLM, BRL, TZS, NGNC, AUDD, EURC

### Screenshots
Only XLM shown in the dropdown
<img width="569" alt="Screenshot 2024-02-19 at 6 49 34 PM" src="https://github.com/pendulum-chain/portal/assets/52105313/4572b2b3-d304-45a6-99e1-826dba987426">
Asset name not visible in asset column
<img width="1188" alt="Screenshot 2024-02-19 at 6 54 47 PM" src="https://github.com/pendulum-chain/portal/assets/52105313/a8feb861-028b-46fe-b101-72b8f1e0fe19">

### Desktop (please complete the following information):

- OS: MacOS
- Browser: Brave, Chrome

prayagd:
Moving this to development, as i see a PR linked here

PR #346, opened 2024-02-19
fix: add Buffer polyfill to window global object

Sharqiewicz:
### What:
In the Spacewalk, experiencing the error:
Error converting currency to stellar asset ReferenceError: Buffer is not defined

### How:

Fixed by implementing `window.Buffer = Buffer;`

Closes: #345


gonzamontiel (approved):
LGTM

Issue #333, opened 2024-01-31
Enable the Spacewalk option from menu on Pendulum network


prayagd:
## Context
For users to access Spacewalk on Pendulum, it should be enabled on the menu

## Requirements
* Enable the Spacewalk menu item and the sub menu items
* Port over the token icons from Amplitude to the Pendulum version in the spacewalk token dropdown
* Add these icons to the Pendulum dashboard

vadaynujra:
Thanks @prayagd, good to know that this is the only ticket required for enabling Spacewalk on Portal for Pendulum.

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/portal/333?planning-poker) with Zenhub @ebma @gonzamontiel

prayagd:
@gonzamontiel Let me know if anything missing with respect to the tokens i.e icons or prices

PR #347, opened 2024-02-21
feat: enable the spacewalk option from menu on pendulum network

Sharqiewicz:
### What:

Enable the Spacewalk option from menu on Pendulum network

### How:

✅  organize the connect function in NodeInformationProvider into smaller functions
✅  enable Spacewalk for Pendulum in the Layout/links
✅  do not display tryConvertCodeToAscii error in the console
✅  removed unused hooks/spacewalk/index.tsx file
✅  Implemented a toast with an error message in Spacewalk > Bridge when Could not establish connection with the bridge
✅  do not display 'undefined' when the token symbol in the bridge could not be loaded
✅  change the name of vaultRegistry hook to useVaultRegistryPallet due to 'use' hook naming convention

Closes: #333


src/hooks/spacewalk/useBridgeSettings.ts
```diff
@@ -29,17 +30,21 @@ function useBridgeSettings(): BridgeSettings {
...
+          const vaultWithRedeemable = data[1]?.find(([id, _]) => id.eq(vaultFromRegistry.id));
+          const extended: ExtendedRegistryVault = vaultFromRegistry;
+          extended.issuableTokens = vaultWithIssuable ? vaultWithIssuable[1] : undefined;
+          extended.redeemableTokens = vaultWithRedeemable ? vaultWithRedeemable[1] : undefined;
+          combinedVaults.push(extended);
+        });
+        setExtendedVaults(combinedVaults);
+      })
+      .catch(() => {
+        toast('Could not establish connection with the bridge...', { type: toast.TYPE.ERROR, toastId: 2138 });
```
ebma, line 46: What's that `toastId` doing?

ebma (approved):
The refactoring looks good to me, nice job 👌 
If you renamed the file for the vault registry pallet to align with hook naming conventions then let's also do it for the others, or is there a reason you chose not to?

src/hooks/spacewalk/useBridgeSettings.ts
ebma: Ah, I see 👍 but the ID was chosen arbitrarily?

ebma: Hmm yes, let's try that. It's not super necessary but makes it clearer what they are used for and how they are assigned. And it's easier to check which numbers are already assigned 👍

Sharqiewicz: @ebma I've added a `showToast.ts` to create an abstraction over toast to make the toast management cleaner. I implemented it for toast messages without any variables. Do you think I should implement it also for toast messages with variables?

ebma: Hmm yes, I think it's better to streamline it, might cause confusion otherwise. How about we just extend your new function with an optional parameter `customMessage?: string` and if this `customMessage` is passed to the function you ignore the `message` contained in `ToastProperties` but rather use that. This way, the caller can still use variables for the message but we don't have to create some complex string replacement logic.

gonzamontiel (approved):
Great work. LGTM

Issue #267, opened 2023-10-13
Unlock unstaked tokens


prayagd:
As a portal user, i should be able to unlock the unstaked tokens from the staking dashboard

## Accpetance criteria
* on /staking
* If the user has unstaked the tokens and it has been > 7 days
* Enable the "Unlock" button
* Clicking on it, should show a pop-up allowing the user to claim unlocked tokens
* If successful show the success message
* If it has been less than 7 days, disable the button and show the amount of days left for unlock on the tool tip. 

### Notes
* Extrinsic to call to unlock the unstaked tokens - `parachainStaking -> unlockUnstaked`
* This flow is only valid after the user unstakes the token from any collator
* The 'Unlock' button should be disabled if there are no tokens to unlock, or the user has not unstaked the tokens
* If there are no tokens to unlock, dont show any tool tip message

### Design flow
Designs titlled with "Unstake unlock" - https://www.figma.com/file/5up5sghcttBogehUq4255i/App?type=design&node-id=2872-2003&mode=design&t=FBdujqSPbvPkSWA1-4

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/portal/267?planning-poker) with Zenhub @ebma @gonzamontiel @Sharqiewicz

PR #356, opened 2024-03-11
feat: 267 unlock unstaked tokens

Sharqiewicz:
### What:

As a portal user, i should be able to unlock the unstaked tokens from the staking dashboard

### How:

✅  If the user has unstaked the tokens and it has been > 7 days - show user the tokens are ready to be unlocked. (Enable the "Unlock" button)
✅  Clicking on it, should show a dialog allowing the user to claim unlocked tokens
✅  If successful show the success message.
✅  If it has been less than 7 days, disable the button and show the amount of days left for unlock on the tool tip.
✅ Implement unified <Dialog/> component
✅ Update all modals to use unified Dialog component

### Comments
Waiting to resolve issue with unlocking all tokens at once - which not in line with designs

Closes: #267


prayagd (changes requested):
Two things
1. The text on the confirmation pop-up is incorrect, It should say "You have successfully unlocked X PEN tokens"
2. The confirmation pop-up has a "Claim" button instead it should be "Ok"
Attached screenshot
image


Otherwise all good!

Sharqiewicz:
@prayagd fixed in the last commit

src/pages/collators/dialogs/unlock/UnlockValidationSchema.ts
```diff
@@ -0,0 +1,13 @@
+import * as Yup from 'yup';
+import { UnlockFormValues } from './UnlockDialog';
+import { transformNumber } from '../../../../helpers/yup';
+
+export function getUnlockValidatiomSchema(balance: number) {
```
ebma, line 5:
```suggestion
export function getUnlockValidationSchema(balance: number) {
```

src/shared/constants.ts
```diff
@@ -3,3 +3,7 @@ export const cacheKeys = {
   accountBalance: 'accountBalance',
   tokenAllowance: 'tokenAllowance',
 };
+
+export const FIFTEEN_MINUTES_IN_MILLISECONDS = 15 * 60 * 1000;
```
ebma, line 7: Maybe it makes more sense to have a constant `MINUTES_IN_MILLISECONDS` and then multiply by `15` in the places where 15 minutes are needed.

ebma (changes requested):
The dialog to unlock the staked tokens still has a textfield that suggests I can change this value although I cannot. IIRC you already raised this issue on Slack @Sharqiewicz, can we address this issue in the same PR here @prayagd?

The two buttons share the same loading indicator, ie. they are showing the indicator at the same time although it doesn't make sense.
image

prayagd:
> The dialog to unlock the staked tokens still has a textfield that suggests I can change this value although I cannot. IIRC you already raised this issue on Slack @Sharqiewicz, can we address this issue in the same PR here @prayagd?
@ebma there is already a temproray fix in the place those numbers could not be edited, though they look like they do. As changing the textfield would be a design change, i would suggest we go ahead with this temporary fix until we get a new design which shows that amount cannot be modified.

Sharqiewicz:
✅ Ready @ebma

src/pages/collators/CollatorRewards/helpers.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+  } else if (status.isFinalized) {
+    setSubmissionPending(false);
+    refreshRewards();
+    if (errors.length === 0) {
+      showToast(ToastMessage.UPDATED_DELEGATOR_REWARDS);
+      setUpdateEnabled(false);
+      setTimeout(() => {
+        setUpdateEnabled(true);
+      }, FIFTEEN_MINUTES_IN_MILLISECONDS);
```
ebma, line 88: I meant to simplify it to this because it's just as readable and we don't end up with having constants for 'TEN_MINUTES_IN_MILLISECONDS', `ELEVEN_MINUTES_IN_MILLISECONDS', yadda yadda 

```suggestion
      }, 15 * MINUTES_IN_MILLISECONDS);
```

src/pages/collators/dialogs/unlock/UnlockValidationSchema.ts
ebma: This typo still exists. Doesn't matter much but we can fix it anyways

ebma (approved):
Nice, LGTM 👍

PR #361, opened 2024-03-13
chore: Configure semantic-release to add release commit on feature branch when PR is closed

Sharqiewicz:
### What:

Configure semantic-release to add release commit on feature branch when PR is closed

### How:

✅ Update Github Actions Workflow
✅ Update semantic-release config to include creation of Github Release

### Comments:

The failures were due to Semantic-Release attempting to push both the "release tag" and the "release commit" directly to a protected branch.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,7 @@ on:
 jobs:
   semantic-release:
     name: Running semantic-release on PR
+    if: ${{ github.event.pull_request.merged == true }}
     runs-on: ubuntu-latest
     strategy:
       matrix:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,15 +1,13 @@
-name: semantic release
+name: semantic-release
 on:
-  pull_request:
+  push:
     branches: [staging]
-    types: [closed]
 
   workflow_dispatch:
 
 jobs:
   semantic-release:
-    name: Running semantic-release on PR
-    if: ${{ github.event.pull_request.merged == true }}
+    name: Running semantic-release
     runs-on: ubuntu-latest
     strategy:
       matrix:
@@ -21,9 +19,9 @@ jobs:
       - name: 🛒 Checkout code
         uses: actions/checkout@v3
         with:
+          persist-credentials: false
           fetch-depth: 0
           token: ${{ secrets.GITHUB_TOKEN }}
-          ref: ${{ github.event.pull_request.head.ref }}
 
       - name: 💿 Setup Nodejs
         uses: actions/setup-node@v3
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,14 +1,14 @@
-name: release
+name: semantic release
 on:
-  push:
-    branches: [staging, develop]
+  pull_request:
+    branches: [staging]
+    types: [closed]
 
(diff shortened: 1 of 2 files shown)


prayagd:
@Sharqiewicz moving this to review as i see you have asked for review

ebma:
Can you elaborate how the changes in this PR fix the problem?

> The failures were due to Semantic-Release attempting to push both the "release tag" and the "release commit" directly to a protected branch.

So we are pushing the release commit to the PR branch before merging it now or what's the idea?

Sharqiewicz:
> Can you elaborate how the changes in this PR fix the problem?
> 
> > The failures were due to Semantic-Release attempting to push both the "release tag" and the "release commit" directly to a protected branch.
> 
> So we are pushing the release commit to the PR branch before merging it now or what's the idea?

@ebma  The idea is to run a semantic release on the feature branch when the PR is merged to staging, so the commit release and tag release are added to the feature branch instead of trying to add them to the protected staging branch. 
So the process looks like this: you click to merge -> release is being added to the feature branch -> branch is being merged with staging. The setup also adds @semantic-release/github to also publish a github release.

ebma:
> The idea is to run semantic release on the feature branch when the PR is merged to staging, so the commit release and tag release are added to the feature branch instead of trying to add them to the protected staging branch.

But the timing here doesn't work, does it? If the workflow is run when/after I merge the PR, then how can the workflow add a commit and tag on the merged branch? We'd need to trigger this action before the PR is closed. Maybe with a bot that has a command to trigger the semantic release and merge the branch right after. Though creating a bot just for this is probably overkill. Please correct me if I'm wrong about the timing.

Sharqiewicz:
> > The idea is to run semantic release on the feature branch when the PR is merged to staging, so the commit release and tag release are added to the feature branch instead of trying to add them to the protected staging branch.
> 
> But the timing here doesn't work, does it? If the workflow is run when/after I merge the PR, then how can the workflow add a commit and tag on the merged branch? We'd need to trigger this action before the PR is closed. Maybe with a bot that has a command to trigger the semantic release and merge the branch right after. Though creating a bot just for this is probably overkill. Please correct me if I'm wrong about the timing.

You're right, the timing doesn't work... I think we can change this Github Action to run on `issue_comment: types: [created]` and add a check `github.event.issue.pull_request && contains(github.event.comment.body, '/release')` and also a check if the comment author is the same as PR author, so if a PR author add a comment `/release` then the semantic-release run, WDYT? @ebma

ebma:
Yeah that's a good idea and could work. We probably should use a different keyword though as just checking if it contains '/release' is a bit too risky, considering there could be URLs that contain this string though we don't want to trigger a new release. We should make it more explicit by checking for 'action semantic:release" or similar.

But taking a step back, if the problem is that semantic release cannot push to a protected branch, can't we just adjust the settings for that branch and make an exception for that action?

ebma:
Let's try changing it as is described [here](https://stackoverflow.com/a/76550826) first. The permissions of the GITHUB_TOKEN and repository should be fine, so let's add the `persist-credentials: false` and then merge to see if it works.

ebma:
Would a new release be triggered once we merge this? Or do we need something else? Just want to make sure that the action is also triggered and tested once we merge this PR.

ebma (approved):

PR #366, opened 2024-03-15
Chore: Add missing v1.3.0 release tag

Sharqiewicz:
### What:

After changing the root branch from `master` to `main` the git tag `v1.3.0` was lost in the history. Semantic-release couldn't find the tag, so it kept creating the `v1.3.0` release as a new release - causing an error. This PR adds the lost git `v1.3.0` tag to the history.


Sharqiewicz:
Outdated

PR #374, opened 2024-03-18
fix: Fix function for filtering assets

### LINKED ISSUE #373: Vault selector dropdown broken on all networks
Currently, the Spacewalk vault selector is broken on all networks and the whole Spacewalk bridging page becomes unresponsive. The console shows the following error:

image

ebma:
Closes #373.


ebma:
Merging this already because the change is so simple.

PR #403, opened 2024-04-10
Fix: Enable stake option for users already delegating to collator

Sharqiewicz:
### What:

Enable stake option for users already delegating to collator

### How:

Fix stake button logic


ebma (approved):
I fixed the condition. Please check again and we merge @Sharqiewicz

PR #407, opened 2024-04-12
fix: handle selectedVault of type number and { amount: number }

Sharqiewicz:
### What:

Error: [big.js] Invalid number
Instead of passing a number to create Big, selectedVault was of type { amount: number } which was causing the error

### How:

✅  handle selectedVault of type number and { amount: number }

### DIFF
--- a/src/pages/bridge/Issue/index.tsx
+++ b/src/pages/bridge/Issue/index.tsx
@@ -194,9 +194,7 @@ function Issue(props: IssueProps): JSX.Element {
           />
           <input type="hidden" {...register('securityDeposit')} />
           <label className="label flex align-center">
-            <span className="text-sm">
-              {`Max issuable: ${nativeToDecimal(issuableTokens || 0).toFixed(2)} ${selectedAsset?.code || ''}`}
-            </span>
+            <span className="text-sm">{`Max issuable: ${maxIssuable.toFixed(2)} ${selectedAsset?.code || ''}`}</span>
           </label>
 
           <FeeBox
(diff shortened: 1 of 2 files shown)


src/pages/bridge/Issue/index.tsx
```diff
@@ -193,9 +195,7 @@ function Issue(props: IssueProps): JSX.Element {
           <input type="hidden" {...register('securityDeposit')} />
           <label className="label flex align-center">
             <span className="text-sm">
-              {`Max issuable: ${nativeToDecimal(selectedVault?.issuableTokens?.toString() || 0).toFixed(2)} ${
-                selectedAsset?.code || ''
-              }`}
+              {`Max issuable: ${nativeToDecimal(issuableTokens || 0).toFixed(2)} ${selectedAsset?.code || ''}`}
```
ebma, line 198: How about we also use the `maxIssuable` variable here?

src/pages/bridge/Redeem/index.tsx
```diff
@@ -159,9 +160,7 @@ function Redeem(props: RedeemProps): JSX.Element {
             }}
           />
           <label className="label flex align-center">
-            <span className="text-sm">{`Max redeemable: ${nativeToDecimal(
-              selectedVault?.redeemableTokens?.toString() || 0,
-            ).toFixed(2)}
+            <span className="text-sm">{`Max redeemable: ${nativeToDecimal(redeemableTokens || 0).toFixed(2)}
```
ebma, line 163: Same here, we can use `maxRedeemable`

ebma (approved):
Nice 👍

Issue #410, opened 2024-04-15
Fetch allowed currencies and native currency from blockchain


Sharqiewicz:
Now Portal is using hardcoded data to show the GET AMPE swap form. ( useBuyout/utils.ts )
We want to use the metadata from the blockchain instead of the hardcoded one.

1. Fetch allowed currencies from the blockchain.
2. Fetch native currency from the blockchain.
3. Handle errors related with undefined data fetched from the blockchain.

Use `api.query.treasuryBuyoutExtension.allowedCurrencies.entries()` to fetch allowed currencies.
Use `api.query.assetRegistry.metadata(...)` to fetch allowed currencies metadata.
Use `api.query.assetRegistry.metadata('Native')` to fetch native currency.

ebma:
Nice ticket, thanks for creating it @Sharqiewicz 👌

PR #411, opened 2024-04-15
Fetch currencies from blockchain instead of using harcoded metadata

Sharqiewicz:
Fetch currencies from blockchain instead of using harcoded metadata 

✅ Fetch allowed currencies from blockchain
✅ Fetch native currency from blockchain
✅ Implement loading /gas
✅ Handle screen if currencies are undefined

Closes: #410


ebma (changes requested):
Nice changes 👍  we could do it like this. However, I would prefer if we make this asset metadata available on a more global level. The reason being that we can use and profit from that metadata not only in the buyout feature but e.g. also in the `usePriceFetcher` hook. Rewriting that logic is out of scope of this PR but we can already prepare a simplified implementation by refactoring here.

What I have in mind:
- We create a hook `useAssetRegistryMetadata()` that is cached, similar to [here](https://github.com/pendulum-chain/portal/blob/a4c117ce1f8e1a7b5a3ab27e3303a6136d595d29/src/shared/useAccountBalance.ts).
- This hook queries and stores the entries of `api.query.assetRegistry.metadata(...)` as well as `api.query.assetRegistry.locationToAssetId()`. 
- The cached response of the `metadata()` query can then be used by the `useBuyout()` hook to map the `CurrencyId`s of the allowed currencies to their respective `Metadata`. We don't need the data of `locationToAssetId()` yet but we will eventually once we implement some feature that allows cross-chain transfers in the portal.

In a follow-up ticket we can then change the implementation of the `usePriceFetcher()` hook to use the information stored in the `additional: {diaKeys: {blockchain, symbol}}` field of each metadata entry to query the prices from chain.

Sharqiewicz:
> Nice changes 👍 we could do it like this. However, I would prefer if we make this asset metadata available on a more global level. The reason being that we can use and profit from that metadata not only in the buyout feature but e.g. also in the `usePriceFetcher` hook. Rewriting that logic is out of scope of this PR but we can already prepare a simplified implementation by refactoring here.
> 
> What I have in mind:
> 
> * We create a hook `useAssetRegistryMetadata()` that is cached, similar to [here](https://github.com/pendulum-chain/portal/blob/a4c117ce1f8e1a7b5a3ab27e3303a6136d595d29/src/shared/useAccountBalance.ts).
> * This hook queries and stores the entries of `api.query.assetRegistry.metadata(...)` as well as `api.query.assetRegistry.locationToAssetId()`.
> * The cached response of the `metadata()` query can then be used by the `useBuyout()` hook to map the `CurrencyId`s of the allowed currencies to their respective `Metadata`. We don't need the data of `locationToAssetId()` yet but we will eventually once we implement some feature that allows cross-chain transfers in the portal.
> 
> In a follow-up ticket we can then change the implementation of the `usePriceFetcher()` hook to use the information stored in the `additional: {diaKeys: {blockchain, symbol}}` field of each metadata entry to query the prices from chain.

Very good idea ✅  I implemented the useAssetRegistryMetadata hook with `api.query.assetRegistry.metadata`. I think we should not implement the `locationToAssetId` now as is not needed (YAGNI rule). I think we can implement the usePriceFetcher update in the `GET PEN` feature PR - WDYT?

src/hooks/useBuyout/index.ts
```diff
@@ -55,11 +55,42 @@ function handleBuyoutError(error: string) {
...
+
+  useEffect(() => {
+    async function fetchAllowedCurrencies() {
+      const allowedCurrencies = [];
+      if (api) {
+        const allowedCurrenciesEntries = await api.query.treasuryBuyoutExtension.allowedCurrencies.entries();
+
+        for (const currency of allowedCurrenciesEntries) {
+          if (currency.length && currency[0] && currency[0].toHuman()) {
+            const currencyXCMId: { XCM: number } = (currency[0].toHuman() as { XCM: number }[])[0];
```
ebma, line 75: This is a bit ugly and we should try to use proper types here. We first need to update them in the pendulum.js package though, so I'm fine if we keep it like this for now.

ebma (approved):
Nice changes 👍 
The hook is not cached now as it would be with the `useQuery` hook [here](https://github.com/pendulum-chain/portal/blob/a4c117ce1f8e1a7b5a3ab27e3303a6136d595d29/src/shared/useAccountBalance.ts) so it will always re-query the data whenever we use it but I think that's negligible. I'm fine if we merge it likes this, good job 👍

Issue #422, opened 2024-04-24
Incorrect unlock period displayed


prayagd:
### Describe the bug

After the user are unlocked after 7 days, the tool tip text displays locked for 7 days which is incorrect

### Reproduction

1. Go to portal.pendulumchain.org/staking
2. unstake the tokens wait for them to unlock, once they are the unlock buttons gets enabled
3. Hover over the tooltip which displays, "Locked for 7 days"

### Expected behavior

The tool tip text should say once the tokens are unlocked as "Tokens unlocked"

### Screenshots
![image](https://github.com/pendulum-chain/portal/assets/52105313/ec9f525b-1f81-49ee-8609-3aca801cc174)

PR #429, opened 2024-04-27
fix Incorrect Unlock Displayed

Sharqiewicz:
### What:
Several days after the start of the process of unstaking tokens - the number of days left doesn't change.

### How:

✅ Instead of parachainStaking.unstaking `toJSON` use `toHuman` - the balance was shown as an address instead of a number of tokens
✅ Fix handling of native (balance from blockchain) - handle i.e. input of `"5,200,000,000"` 
✅ Delete the doubled file (CollatorRewards.tsx)

Closes: #422


ebma (approved):
Nice, looks good to me 👌

Issue #417, opened 2024-04-19
Update error message for insufficient security deposit


annatekl:
### Context
Currently, when the user doesn't have enough PEN to cover the requested security deposit, they see the message: 'You don't have enough balance to cover the deposit fees. 

### Requirement
- [ ] the ticket error message should be: `Insufficient PEN to cover the security deposit.`

<img width="699" alt="Screenshot 2024-04-19 at 13 00 09" src="https://github.com/pendulum-chain/portal/assets/138758794/11c300e5-1612-44cd-a2b6-c2c21518bf0c">

### Note
* Make the same changes to the Amplitude version and replace the PEN as AMPE in the text

vadaynujra:
@annatekl what is pending on this one to be moved to Ready?

annatekl:
this can be moved to ready @vadaynujra

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/portal/417?planning-poker) with Zenhub @ebma @Sharqiewicz

PR #443, opened 2024-05-10
Update error message for insufficient security deposit

Sharqiewicz:
### What:

Change Spacewalk security deposit error
Change Spacewalk tokenSymbol

Closes: #417


ebma (commented):
The error message looks good but I noticed that for this particular error, the 'Bridge' button does not show as disabled while for any other error regarding the schema, it disables. Any ideas @Sharqiewicz?

image

Sharqiewicz:
> The error message looks good but I noticed that for this particular error, the 'Bridge' button does not show as disabled while for any other error regarding the schema, it disables. Any ideas @Sharqiewicz?
> 
> image

✅ Very good catch @ebma 
The button was implemented with `disabled={!!formState.errors.amount}` and it is not covering all of the errors that can appear in the form. I changed it to `disabled={!isEmpty(formState.errors}` and also implemented to trigger validation for every onBlur event.

src/pages/bridge/Issue/index.tsx
```diff
@@ -158,8 +172,10 @@ function Issue(props: IssueProps): JSX.Element {
 
   useMemo(() => {
     setValue('securityDeposit', amount * issueGriefingCollateral.toNumber());
-  }, [amount, issueGriefingCollateral, setValue]);
+    trigger('securityDeposit');
+  }, [amount, issueGriefingCollateral, setValue, trigger]);
 
+  console.log('formState.errors', formState.errors);
```
ebma, line 178: Nit: remove

src/pages/bridge/Issue/index.tsx
```diff
@@ -53,8 +66,9 @@ function Issue(props: IssueProps): JSX.Element {
 
   const maxIssuable = nativeToDecimal(issuableTokens || 0).toNumber();
 
-  const { handleSubmit, watch, register, formState, setValue } = useForm<IssueFormValues>({
-    resolver: yupResolver(getIssueValidationSchema(maxIssuable, parseFloat(balance || '0.0'))),
+  const { handleSubmit, watch, register, formState, setValue, trigger } = useForm<IssueFormValues>({
+    resolver: yupResolver(getIssueValidationSchema(maxIssuable, parseFloat(balance || '0.0'), tokenSymbol)),
+    mode: 'onBlur',
```
ebma, line 71: I think `onBlur` is not often enough. Let's trigger the validation as often as possible, ie every time the input changes. Otherwise, I see errors although they should be gone already.

image

Issue #377, opened 2024-03-19
Close the older pop-up when successfully staked


prayagd:
## Context
When a user is able to successfully stake, portal shows a pop-up about the confirmation about the stake but the earlier pop-up for adding the inputs and staking does get closed. 

## Requirement
* Close the earlier pop-up while staking/unstaking is successful
* Only close if it is succesful, show it if staking/unstaking unsuccessful
* To be implemented on both chains i.e Pendulum and Amplitude

## Reference
![image](https://github.com/pendulum-chain/portal/assets/52105313/607c5bd9-222d-462c-bf39-c60f188f92b4)

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/portal/377?planning-poker) with Zenhub @ebma @Sharqiewicz

PR #444, opened 2024-05-13
Close the older pop-up when successfully staked

Sharqiewicz:
### What:

When a user is able to successfully stake, portal shows a pop-up about the confirmation about the stake but the earlier pop-up for adding the inputs and staking does get closed.

### Comments:

Closes: #377


Sharqiewicz:
This changes will be included in #450

Issue #445, opened 2024-05-14
Screen freezes when manually selecting vault for EURC on Spacewalk


vadaynujra:
Which user interface was the bug observed on?
Portal Spacewalk Bridge settings wheel crashes the interaction

What steps did you take?
Portal
Bridge
Change Asset to EURC
Settings wheel
Check manually select vault
UI freezes nothing else possible to select click or get out of the menu

What was the outcome you expected?
Not freezing, or being able to revert everything, not need to refresh page

What was the outcome you observed?
UI freezes nothing else possible to select click or get out of the menu
Screenshot of the bug / error message
https://satoshipay.slack.com/files/UA3GK5K9P/F073A8FJ0CD/bridge_settings_bug.jpg

Account address that is facing the bug/issue
no wallet or account was connected

Criticality of the bug
Medium

What device were you on?
Desktop

Device model
Macbook Pro 14 M1Pro

OS
MacOS

Browser and its version
Chrome

Additional information regarding the bug
Seems that there is just one, so actually no option, maybe just adjust or avoid the selection (manually the vault)

ebma:
I can confirm and reproduce the bug. Seems to happen with all non-XLM currencies though. Possibly because the max issuable amount of the XLM vault is 0 at the moment, so this doesn't cause any problems.
@Sharqiewicz I took a picture of the console logs, can you please look into this?
<img width="1525" alt="image" src="https://github.com/pendulum-chain/portal/assets/6690623/31bac98f-185f-4ec1-85f6-37044b519660">

PR #457, opened 2024-05-20
Screen freezes when manually selecting vault for EURC on Spacewalk

Sharqiewicz:
### What:

Which user interface was the bug observed on?
Portal Spacewalk Bridge settings wheel crashes the interaction

What steps did you take?
Portal
Bridge
Change Asset to EURC
Settings wheel
Check manually select vault
UI freezes nothing else possible to select click or get out of the menu

### How:

The vault object has nested `amount` property that stores the required value. We passed the object `{ amount: "number-value"}` instead of a valid numeric value to the Big constructor.

Closes: #445

### DIFF
--- a/src/components/Selector/VaultSelector.tsx
+++ b/src/components/Selector/VaultSelector.tsx
@@ -43,7 +43,7 @@ function VaultSelector(props: VaultSelectorProps): JSX.Element {
               </span>
               {showMaxTokensFor && (
                 <span className="content-end">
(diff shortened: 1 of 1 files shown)


ebma (approved):
Nice catch 👍

prayagd:
Screenshot 2024-05-24 at 12 12 19 PM
@Sharqiewicz works for EURC, tried for other assets. the screen freezed for me when trying the XLM asset manual vault selection

PR #477, opened 2024-05-29
Don't show empty balances when account not connected

### LINKED ISSUE #469: Show BSP and swap balance 0 when wallet not connected
### Describe the bug
Currently when the wallet is not connected to the portal, the swap UI and the BSP UI loads continuously showing a animation

### Reproduction

1. Go to https://portal.pendulumchain.org/foucoco/nabla/backstop-pools
2. See the loading animation
3. Same applies for the swap UI, the balances component also keeps loading

### Expected behavior
Nothing should be displayed as the wallet is not connected to the Portal UI

### Screenshots
Screenshot 2024-05-28 at 4 02 49 PM
Screenshot 2024-05-28 at 4 03 01 PM

### Desktop (please complete the following information):

- OS: MacOS
- Browser: Chrome, Firefox

ebma:
I replaced all related loaders I could find with just 'nothing' expect for the Backstop Pool screen where 'My pool balance' now shows `0` if no account is connected.

Closes #469.


prayagd (approved):
All good @ebma with the swap UI, can we remove the "0" also from the backstop pool? and then go ahead and merge.

ebma:
I left the 'My pool balance: 0' at the backstop pool because I thought it might look weird because it's so empty otherwise. But I can remove that too, no problem.

prayagd:
@ebma the text "My pool balance" also went away

ebma:
Wait, I thought that's how you wanted it to look like? 😅  At least that's exactly what I changed it to. Did you expect something else? I can also change it again.

Sharqiewicz (approved):
@ebma perfect ✅

prayagd:
> "My pool balance"

@ebma oh sorry for the confusion. just show this text back, that's it

ebma:
Out of the three options we have:
1) Show 'My pool balance: 0'
2) Don't show 'My pool balance' at all
3) Only show 'My pool balance' but without an actual number after it

Isn't option 3) the most confusing? 😅

vadaynujra:
If there is nothing the user can do on that page, without connecting their wallet, then that's where we should be sending them (and so should be what they see).

References from other DEXes

HydraDX's Wallet Page:

image

Pablo's portfolio view

image

Both dApps:

- fill up the screen to tell the user they should connect their wallet - on the lines of @TorstenStueber's suggestion
- while still keeping placeholders of what the user should expect to see, once they do connect their wallet - I think this is what @prayagd wanted to cater to with the suggestion of keeping the `My Pool Balance`

Specifically on the options:

- Show 'My pool balance: 0' - this would be incorrect, as 0 is also a number and may or may not be true
- Don't show 'My pool balance' at all - user wouldn't know that they can see their balance on this screen once they connect
- Only show 'My pool balance' but without an actual number after it - this is totally fine as long as it is clear to the user that this is a placeholder, which will have a true value one they connect their wallet.

Since my suggested solution might require some more work, can the immediate solution just be that the animation is made static, everything else remaining the same? See HydraDX's [wallet view](https://app.hydradx.io/wallet/assets) without connecting wallet for reference.

TorstenStueber:
I suggest: remove "My pool balance" as @ebma did and just add a "Connect Wallet" button instead. Don't we have a component for this and adding such a button is a trivial change?

ebma:
Thanks for providing some references. Personally, I think it's fine the way it is now with not showing anything related to balances as long as no wallet is connected. I can add the change @TorstenStueber suggested and instead make the button for depositing/swapping prompt to connect a wallet account instead. The same is already done when you go to the Spacewalk issue/redeem dialog without being connected to a wallet.

vadaynujra:
Sounds good, so both buttons become Connect Wallet, when the user connects wallet, the My Pool Balance appears with the correct amount.

prayagd:
@ebma looks good we can merge, checked all the buttons and all the connect wallet buttons work and open the pop-up

PR #490, opened 2024-06-03
fix nabla modals

Sharqiewicz:
### What:

After a user successfully deposits asset into the swap pool, the user still sees the Deposit pop-up after the successful confirmation pop-up is closed

### How:

✅ Use reusable <Dialog> component for Nabla
✅ Fix onClose for Nabla dialogs


ebma (approved):
Nice, looks good to me 👍

PR #492, opened 2024-06-03
fix lottie animation stops playing onclick

Sharqiewicz:
### What:

When you hover your cursor over “Stellar Bridge” in the navigation, the Lottie animation starts playing, however, if you click on it while the animation is playing, the animation stops and cannot continue.

### How:

✅ Add `isClickToPauseDisabled={true}` to the Lottie component
✅ Add `createLottieOptions` function to make the code less repetitive

⚠️ Note
In the console this Lottie error appears:
`Failed prop type: Invalid prop 'style' of type 'object' supplied to 'Lottie2', expected 'string'`
The error appears since the day we implemented it, unfortunately the developer of the react-lottie library implemented the types incorrectly and the styles of the `<Lottie>` component cannot be implemented without this error


ebma (approved):
Nice, looks good to me 👍 and thanks for the note about the error in the console, I saw it and was wondering a few days ago.

Issue #489, opened 2024-06-03
Add space between the token icon and the name on Forex AMM token dropdown


prayagd:
### Describe the bug

On the swap UI of the Forex AMM in the token dropdown pop-up, there is no space between the token name and the token icon.

### Reproduction

1. Go to /nabla/swap
2. Click on the token selection dropdown
3. See there is no space between the token name and the icon

### Expected behavior

There should be some space between the icon and the name

### Screenshots

![image](https://github.com/pendulum-chain/portal/assets/52105313/c155b125-279e-4dce-a0d8-31308e151d4b)

### Desktop (please complete the following information):

- OS: MacOS
- Browser: Firefox

Issue #478, opened 2024-05-30
Remove 50% and Max from the Backstop pool withdraw UI


prayagd:
## Context
Currently the max values are not calculated properly in the Backstop pool(BSP) withdraw UI

## Requirement
* Remove the 50% and Max options from 
    * backstop pool withdrawal (return backstop pool LP tokens and receive backstop pool tokens)
    * swap pool withdrawal (return swap pool LP tokens and receive swap pool tokens)
    * backstop redeem (return swap pool LP tokens and receive backstop pool tokens)
    * backstop burn

## Note
This is both for the Pendulum and Foucoco deployment

## Reference
<img width="557" alt="Screenshot 2024-05-30 at 12 48 54 PM" src="https://github.com/pendulum-chain/portal/assets/52105313/65f8d68f-ded8-4427-b0ca-a1789726c58f">

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/portal/478?planning-poker) with Zenhub @ebma @Sharqiewicz @TorstenStueber

ebma:
What does 'not calculated properly' mean? Can you give an example when it fails? Is removing the buttons the best idea to solve this though?

TorstenStueber:
The way it was defined in the code (and I did not change this anymore when I took it over): the max button selects the maximum amount of tokens you can deposit, i.e., 100% of your LP tokens. Similar for the 50% button.

However, that is not the actual maximum amount possible. In some situations you are not able to return all LP tokens because of some limits or because there are not enough tokens in the pool to return.

Here the argument is that the user would expect that the Max button does not select 100% of your LP tokens but the exact maximum of tokens you are able to return in this situation. 

This kind of problem happens for four operations:
* backstop pool withdrawal (return backstop pool LP tokens and receive backstop pool tokens)
* swap pool withdrawal (return swap pool LP tokens and receive swap pool tokens)
* backstop redeem (return swap pool LP tokens and receive backstop pool tokens)
* backstop burn (return backstop pool LP tokens and receive swap pool tokens)

There is no function in the smart contracts to calculate these maximum values (and we can't add them, otherwise they become too big for pallet contracts).

That means that we would need to do the calculation of the current maximum amount in the UI. This is not trivial and I wanted to leave it out in the initial version of the Nabla UI.

TorstenStueber:
[...]

PR #496, opened 2024-06-10
Remove 50% and Max from the Backstop pool withdraw UI

Sharqiewicz:
### What:

Remove 50% and Max from the Backstop pool withdraw UI

✅  Change color of Nabla dialog inputs
✅  Fix Nabla loading buttons styles
✅  Remove 50% and Max buttons
✅  Refactor Nabla modals to use reusable dialog components
✅  Add a space between Nabla Pool dialog asset and asset name ( Closes: #489 )

Closes: #478


ebma (changes requested):
I can confirm that the change for https://github.com/pendulum-chain/portal/issues/489 is as expected.

For the '50%' and 'MAX' button, ticket https://github.com/pendulum-chain/portal/issues/478 specified that these should only be removed when withdrawing, not for depositing. So we should not remove them altogether but just hide them in these cases.

Sharqiewicz:
@ebma I have corrected what you asked for. Now when you click 50% or Max you may find that too many decimals are shown. This is corrected in PR #506 so there is no point in implementing it in this PR

src/components/Form/From/AvailableActions/index.tsx
```diff
@@ -1,13 +1,14 @@
 interface AvailableActionsProps {
   max?: number;
   setValue?: (n: number) => void;
+  hideAvailable?: boolean;
```
ebma, line 4: I was confused when first reading it, let's maybe be more specific
```suggestion
  hideAvailableBalance?: boolean;
```

ebma (commented):
Works for swapping but the actions are now also missing for the deposit screen of the backstop pool which they should not.

image

Sharqiewicz:
@ebma I've added for Backstop Pool Deposit and I changed the naming

ebma (approved):
Cool, works now 👍

PR #512, opened 2024-06-26
show Insufficient balance tooltip only when user is logged in

Sharqiewicz:
✅ show Insufficient balance tooltip only when user is logged in


Sharqiewicz:
Return to the issue when new wallet connection is implemented and merged in this PR: #498

Sharqiewicz:
Return to the issue when new wallet connection is implemented and merged in this PR: #498

PR #515, opened 2024-07-01
Fix parsing of amounts in vault selector

### LINKED ISSUE #508: Manual selection of Vault - inconsistent across currencies
The system behaves inconsistently for different currencies when the user manually selects the vault.

### Reproduction (bug for XLM)

1. Go to portal (choose the bridge to Pendulum or to Amplitude)
2. Choose XLM asset and click on settings
3. Click on 'Manually select vault'
4. Observe: The system becomes unresponsive.

### Screenshots
Screenshot 2024-06-26 at 08 40 17

### Reproduction (bug for EURC)

1. Go to portal (choose the bridge to Pendulum or to Amplitude)
2. Choose EURC asset and click on settings
3. Click on 'Manually select vault'
4. Choose the vault and click on ok
5. Attempt to change the asset to another asset from the list.
6. Observe: The change is not possible, EURC remains as the chosen asset

#### Expected behaviour:
- For XLM: The system should remain responsive and allow the user to proceed with their actions.
- For EURC: The user should be able to successfully change the asset after manually selecting a vault.

ebma:
- [x] Fix alignment of amount text in dropdown item of manual vault selection (there was no space between vault ID and amount)
- [x] Make parsing more resilient ie catch errors. 
- [x] Fix bug where always the issuable amount was used, even on the manual vault selection for redeeming

The problem that 'broke' the feature in the first place is that the RPC call for `getVaultsWithIssuableTokens()` for some reason returns an empty array on Pendulum at the moment.

Closes #508.

### DIFF
--- a/src/components/Selector/VaultSelector.tsx
+++ b/src/components/Selector/VaultSelector.tsx
@@ -1,5 +1,6 @@
 import { ChevronDownIcon } from '@heroicons/react/20/solid';
 import { Button, Dropdown } from 'react-daisyui';
+import { BridgeDirection } from '../../pages/spacewalk/bridge';
 import { convertCurrencyToStellarAsset } from '../../helpers/spacewalk';
 import { ExtendedRegistryVault } from '../../hooks/spacewalk/useVaultRegistryPallet';
 import { nativeToDecimal } from '../../shared/parseNumbers/metric';
@@ -8,7 +9,7 @@ import { PublicKey } from '../PublicKey';
 interface VaultSelectorProps {
   vaults: ExtendedRegistryVault[];
   selectedVault?: ExtendedRegistryVault;
-  bridgeDirection?: 'issue' | 'redeem';
+  bridgeDirection?: BridgeDirection;
   onChange: (vault: ExtendedRegistryVault) => void;
 }
 
--- a/src/components/Selector/VaultSelector.tsx
+++ b/src/components/Selector/VaultSelector.tsx
@@ -8,12 +8,27 @@ import { PublicKey } from '../PublicKey';
 interface VaultSelectorProps {
   vaults: ExtendedRegistryVault[];
   selectedVault?: ExtendedRegistryVault;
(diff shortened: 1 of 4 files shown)


src/components/Selector/VaultSelector.tsx
```diff
@@ -8,12 +8,27 @@ import { PublicKey } from '../PublicKey';
 interface VaultSelectorProps {
   vaults: ExtendedRegistryVault[];
   selectedVault?: ExtendedRegistryVault;
-  showMaxTokensFor?: 'issuableTokens' | 'redeemableTokens';
+  bridgeDirection?: 'issue' | 'redeem';
```
Sharqiewicz, line 11: Great change, it is definitely more clear right now ✅ 💪 
What do you think about creating a separate type e.g:
`type BridgeDirection = 'issue' | 'redeem'`

and use it also in the `getMaxTokensForVault` type

src/pages/spacewalk/bridge/index.tsx
```diff
@@ -52,10 +52,16 @@ function Bridge(): JSX.Element | null {
     onClick: () => setTabValue(index),
   });
 
+  const bridgeDirection = tabValue === BridgeTabs.Issue ? 'issue' : 'redeem';
```
Sharqiewicz, line 55: As we use 'issue' and 'redeem' strings several times across all of these files, what about we create an enum `enum BridgeDirections {
ISSUE,
REDEEM
}`

and the use it for the `type BridgeDirection`?

Sharqiewicz (commented):
✅ Great changes @ebma

src/components/Selector/VaultSelector.tsx
ebma: Sure 👍  I created that enum and put it into the `src/pages/spacewalk/bridge/index.tsx` file. Let me know if you think that works or if there is a better location.

Sharqiewicz (approved):
Perfect ✅

PR #516, opened 2024-07-01
Trigger validation in redeem immediately and improve `disabled` field

### LINKED ISSUE #510: User Not Prompted to Enter a Positive Number for Assets Bridged Back to Stellar
#### Context
Currently, users can input any amount, including zero, for assets being bridged back to Stellar. However, when zero is entered, the user does not receive an error message prompting them to enter a positive number, similar to the behaviour when bridging assets to Pendulum.

### Describe the feature you'd like:
When a user inputs zero or a negative number for assets to be bridged back to Stellar, an error message should display, stating: "You need to enter a positive number."

ebma:
Triggers the validation for the redeem form immediately. Aligns the `disabled` prop for the 'Bridge' button to the issue form.
Closes #510.

### DIFF
--- a/src/pages/spacewalk/bridge/Redeem/index.tsx
+++ b/src/pages/spacewalk/bridge/Redeem/index.tsx
@@ -1,5 +1,6 @@
 import { yupResolver } from '@hookform/resolvers/yup';
 import Big from 'big.js';
+import { isEmpty } from 'lodash';
 import { useEffect } from 'preact/compat';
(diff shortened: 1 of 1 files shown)


Sharqiewicz (approved):
@ebma Great changes ✅ I've added a commit which removes unused variables and replace `any` with `Signer` type from polkadot

ebma:
Cool, thanks for cleaning it up 🙏

PR #524, opened 2024-07-11
GET AMPE and GET PEN features not working on portal

Sharqiewicz:
### What:

GET AMPE and GET PEN features not working on portal

### How:

Data format returned by `api.query.assetRegistry.metadata` was changed so I adjusted the data comparison. 
Now from both allowedCurrencies and assetRegistryMetadata currencies data is retrieved by `x.currencyId.toHuman()`

Also I replaced `React.FC` with just `FC` from `preact/compat`


ebma (approved):
Nice, works 👍

PR #525, opened 2024-07-18
495 enable circle eurcs as spacewalk asset on the portal on pendulum

### LINKED ISSUE #495: Enable Circle EURC as Spacewalk asset on Pendulum
#### Context
We are planning to add Circle EURC to Spacewalk assets. We need to enable asset on the portal once the [root transaction](https://app.zenhub.com/workspaces/pendulum-6408a0365cda59c57d39bee5/issues/gh/pendulum-chain/tasks/290) is completed.

#### Requirements
- [ ] Enable Circle EURC as Spacewalk asset on the portal
- [ ] Add Circle EURC icon to the Circle EURC asset 
- [ ] Add Circle EURC icon to the dashboard

ebma:
Refactors the logic to get an icon so that it checks using `includes` and not for perfect matches and adds an additional check to decide between the mEURC and cEURC icons.

With these changes, the Circle EURC icon should show properly in the dashboard as well as the asset selector in the Spacewalk menus. We don't care about Nabla yet as that's a special use case where we don't need to have a perfect solution just yet.

Closes #495.


Sharqiewicz (approved):
LGTM 🚀  ✅

ebma:
@Sharqiewicz I noticed that all EURC icons are now showing the Circle EURc though I was not yet able to pin down why. This happens in the dashboard as well as the asset selector of the spacewalk dialogs. Very weird.
image

Sharqiewicz:
> @Sharqiewicz I noticed that all EURC icons are now showing the Circle EURc though I was not yet able to pin down why. This happens in the dashboard as well as the asset selector of the spacewalk dialogs. Very weird. image

No asset matches the Mykobo issuer address, I am resolving it

ebma:
I also thought that something was wrong with that match but actually, if you add plenty of log statements you will see that sometimes, it is able to match it. I think the `issuer` is sometimes not getting passed into the function and that's why it will return the cEURC icon by default. 

I was not able to find the places where the issuer is not passed though.

Sharqiewicz:
@ebma The issue was that we were invoking `getAssetIcon` two times, first for `getPolkadotAssetIcon` and then for `getStellarAssetIcon`.  `getPolkadotAssetIcon` had always value of cEURC and `getStellarAssetIcon` had the value of mEURC. The return condition was `polkadotIcon || stellarIcon || defaultIcon;` so it was always returning cEURC.

ebma:
Ahh I see, thanks a lot for debugging and fixing this @Sharqiewicz! 🙏 I'll merge it then.

PR #527, opened 2024-07-22
improve code readability

Sharqiewicz:
✅ Unify AssetItem object (now it uses AssetItem interface, it already has name - so all of the AssetItem helper functions are private, and all the needed properties are already in the object)

✅ Change the structure of AssetIcons to avoid repetition

### DIFF
--- a/src/shared/AssetIcons.tsx
+++ b/src/shared/AssetIcons.tsx
@@ -18,6 +18,7 @@ import DefaultIcon from '../assets/coins/placeholder.png';
 // The public key of Mykobo's EURC issuer account
 const MYKOBO_ISSUER = 'GAQRF3UGHBT6JYQZ7YSUYCIYWAF4T2SAA5237Q5LIQYJOHHFAWDXZ7NM';
 
+// Maps all supported Stellar assets to icons. The EURC tokens are handled separately in `getAssetIcon()`
 const stellarAssets = [
   { code: 'AUDD', icon: AUDD },
   { code: 'BRL', icon: BRL },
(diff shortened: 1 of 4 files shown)


src/shared/AssetIcons.tsx
```diff
@@ -18,52 +18,47 @@ import DefaultIcon from '../assets/coins/placeholder.png';
...
+  { code: 'DOT', icon: DOT },
+  { code: 'AMPE', icon: AMPE },
+  { code: 'KSM', icon: KSM },
+  { code: 'USDT', icon: USDT },
+  { code: 'GLMR', icon: GLMR },
+];
+
+const getAssetIcon = (assetCode: string, assetIssuer: string, assets: { code: string; icon: string }[]) => {
+  if (assetCode.includes('EURC')) {
+    return assetIssuer === MYKOBO_ISSUER ? mEURC : cEURC;
```
ebma, line 41: I think it can be a bit confusing for a reader, that the definition of `stellarAssets` above doesn't contain EURC.

How about we at least add a comment above `stellarAssets` in l.20 
```
// Maps all supported Stellar assets to icons. The EURC tokens are handled separately in `getAssetIcon()`
```
or similar?

ebma: This was the reason why I included everything in one function, though the very long `if-else` branches are very ugly, that's for sure. So thanks for this improvement 🙏

ebma (approved):
Nice improvements, thanks @Sharqiewicz 👍 
Let's add a comment and we merge.

src/shared/AssetIcons.tsx
Sharqiewicz: Good idea @ebma Let's do it

Sharqiewicz: What do you think about this? @ebma 

```
// Maps all supported Stellar assets to icons. The EURC tokens are handled separately in `getAssetIcon()`
const stellarAssets = [
  { code: 'AUDD', icon: AUDD },
  { code: 'BRL', icon: BRL },
  { code: 'NGNC', icon: NGNC },
  { code: 'TZS', icon: TZS },
  { code: 'USDC', icon: USDC },
  { code: 'XLM', icon: XLM },
];


const handleSpecialAsset = (assetCode: string, assetIssuer: string) => {
  // The EURC tokens are handled separately because they can be issued by multiple issuers
  if (assetCode.includes('EURC')) {
    return assetIssuer === MYKOBO_ISSUER ? mEURC : cEURC;
  }
};

const getAssetIcon = (assetCode: string, assetIssuer: string, assets: { code: string; icon: string }[]) => {
  const specialAsset = handleSpecialAsset(assetCode, assetIssuer)
  if(specialAsset) return specialAsset;

  const asset = assets.find((asset) => assetCode.includes(asset.code));

  return asset?.icon || undefined;
};
```

ebma: I love it, let's do it that way 👍

PR #542, opened 2024-08-28
fetch assets from batching server

Sharqiewicz:
🦈
---

- [x] The filters for the column should work for only two combinations i.e descending to ascending and vice versa and not other combination, currently it works for three combination where the third combination is unclear
- [x] Add a default filter to the portal dashboard based on the USD value of the tokens, from descending to ascending
- [x] Add prices for all the portal assets (Implemented fetching prices from the batching server)
- [x] Fix styles of the tables

### Local Development 
The proxy server is implemented to bypass CORS-related issues when running the portal locally.

Unlike the browser, the proxy server does not trigger a preflight request, which is initiated by the browser to check permissions for cross-origin resource sharing (CORS). The preflight request can block communication between the portal and the batching server, causing interruptions in the connection.


Sharqiewicz:
@pendulum-chain/devs Ready for review. The prices won't work on deploy preview - CORS policy of the browser is going to block communication between the batching server and the portal.

In order to test it you have to run it locally, with local proxy server setup. 
If you want to fetch prices locally, you can use the proxy server available at [pendulum-tools](https://github.com/pendulum-chain/pendulum-tools) (Currently PR in review). Change url in
`src/hooks/usePriceFetcher.ts` file to `http://localhost:3000`

CORS policy of the batching-server is set to `portal.pendulumchain.org` so the production price fetching is going to work.

src/hooks/usePriceFetcher.ts
```diff
@@ -20,68 +20,58 @@ function diaKeysToString(diaKeys: DiaKeys) {
...
+  const diaKeys = allAssetsMetadata.map((asset) => asset.metadata.additional.diaKeys);
 
-  useEffect(() => {
-    if (!api) return;
-
-    const fetchPrices = async () => {
-      const allPrices = await api.query.diaOracleModule.coinInfosMap.entries();
+  const getPriceFromBatchingServer = useCallback(async () => {
+    try {
+      const response = await fetch('https://batching-server.pendulumchain.tech/currencies', {
```
ebma, line 30: Let's maybe move the URL into some other file where we define constants.

ebma (approved):
I tested this locally with a local batching server (with CORS enabled) and the proxy server and I can confirm the changes work. Great job @Sharqiewicz 👍 Let's wait with merging until the new batching server is redeployed.

src/hooks/usePriceFetcher.ts
Sharqiewicz: good idea ✅

Sharqiewicz: @ebma Implemented ✅

ebma: Thanks 👍

ebma:
Will do. This will probably happen on Monday, see the related change request [here](https://github.com/pendulum-chain/tasks/issues/394).

PR #543, opened 2024-08-29
Fix redirect to AlchemyPay

### LINKED ISSUE #536: Make new changes to the PEN onramp
## Context
Alcehmy pay which is our offramp has made some changes to their api for security reasons, we need to update them as the have also impacted our default setup of showing PEN as a onramp option

## To-do
* Upgrade to a new signing method, sharing all the necessary links below
* When redirect from the portal, it should show PEN as offramp option and the fiat currency as per geographic location (In think so this is taken care by the api)

## Notes
- Page Integration Signing Instructions: (https://alchemypay.readme.io/v4.0.2/docs/ramp-signature-description)
- API Integration Signing Instructions: (https://alchemypay.readme.io/v4.0.2/docs/api-sign)
- Webhook Signing Instructions: (https://alchemypay.readme.io/v4.0.2/docs/webhook-signature)
- The new signing method requires the merchantOrderNo parameter to be mandatory and unique (non-repeating).
- The existing signing method will no longer be supported after August 15, 2024.

ebma:
- [x] Creates a new netlify function that has access to the required parameters to sign the query parameters for showing PEN by default on the AlchemyPay dashboard.
- [x] Refactor code to allow for asynchronous generation of navigation links 

Closes #536 and closes https://github.com/pendulum-chain/tasks/issues/382

### DIFF
--- a/src/components/Layout/Nav.tsx
+++ b/src/components/Layout/Nav.tsx
@@ -1,9 +1,9 @@
-import { memo, useMemo, useState } from 'preact/compat';
+import { memo, useEffect, useMemo, useState } from 'preact/compat';
 import { NavLink, useLocation } from 'react-router-dom';
 
 import { useGlobalState } from '../../GlobalStateProvider';
 import useBoolean from '../../hooks/useBoolean';
-import { LinkItem, links } from './links';
+import { createLinks, LinkItem } from './links';
 import { NavCollapseButtonContent } from './NavCollapseButtonContent';
 
 const CollapseMenu = ({
@@ -83,14 +83,19 @@ const Nav = memo(({ onClick }: NavProps) => {
   const state = useGlobalState();
 
   const [isPlaying, setIsPlaying] = useState(false);
+  const [links, setLinks] = useState<LinkItem[]>([]);
 
   const handleMouseEnter = () => {
     setIsPlaying(true);
   };
 
+  useEffect(() => {
+    createLinks(state.tenantName).then((links) => setLinks(links));
+  }, [state.tenantName]);
+
   return (
     <nav>
-      {links(state).map((item, i) => {
+      {links.map((item, i) => {
         if (item.hidden) return;
         return item.submenu ? (
           <div onMouseEnter={handleMouseEnter} className="my-2.5">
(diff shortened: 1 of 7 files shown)


src/components/Layout/Nav.tsx
```diff
@@ -83,14 +83,19 @@ const Nav = memo(({ onClick }: NavProps) => {
...
 
   const [isPlaying, setIsPlaying] = useState(false);
+  const [links, setLinks] = useState<LinkItem[]>([]);
 
   const handleMouseEnter = () => {
     setIsPlaying(true);
   };
 
+  useEffect(() => {
+    createLinks(state.tenantName).then((links) => setLinks(links));
```
TorstenStueber, line 93: I assume that the creation of the links, which entails a call to the backend, would usually be quite fast so that the menu will be shown shortly after the app launch.

However, could it be that it takes longer than a second or more? In that case maybe we can show the menu right from the beginning and either a) leave out the "Buy PEN" entry and only show once the link has loaded or b) show the "Buy PEN" entry right away but disable it as long as the link is not available.

TorstenStueber (commented):
I like the solution with the Netlify function. Just one remark about making the menu available earlier in case the function loads or works slowly... (at least for AWS Lambda this is the case after some inactivity).

src/components/Layout/Nav.tsx
ebma: Thanks for the comment. I changed it so that the function is not async anymore but returns a 'default' array of LinkItems and a promise that resolves to contain the actual fully-prepared LinkItems.

TorstenStueber: I am not sure whether I am reading this right: the [initial link](https://github.com/pendulum-chain/portal/pull/543/files/baaee22c3adad97d21cf4c2d12d3284a950d6429..200787c43a223149b15fde6f2b3ae630b84ab68e#diff-0156eb8485cbc4ab310a91b31d411940c81a71b430bcc84a0974a800765a94c6R145) will be `config.alchemyPay.prodUrl` and will then be replaced by the correct link once the promise resolved. Does that initial link (which is defined to be `https://ramp.alchemypay.org`) make sense to click on?

ebma: Yes, the only difference between `https://ramp.alchemypay.org` and the link we generate is that on the proper link, you will see 'PEN' prefilled in the field for the receiving currency. The rest is the same.

TorstenStueber: Perfect. Thanks for the change.

TorstenStueber (approved):

PR #547, opened 2024-09-02
update yarn, update readme.md on corepack

Sharqiewicz:
Problem:

The `packageManager` field in `package.json` specifies the package manager and its version for the project. However, it doesn't enforce the use of this package manager on its own.

What to do?

- Update Yarn version
- Add the note below to readme.md:

```
Enable Corepack by executing the command `corepack enable`. 
Corepack is included by default in Node.js, it manages the Yarn version specified in the `packageManager` field.

**Important:** Modern [Yarn](https://yarnpkg.com/getting-started/install) releases should not be installed globally or via npm - use Corepack instead.
```


Sharqiewicz:
@pendulum-chain/devs Ready for review. Minor change

PR #548, opened 2024-09-02
hide Insufficient Balance Popup if balance is not ready

Sharqiewicz:
## 🦈 
When I am loading the portal (dashboard), it shows me a warning about insufficient funds to execute transactions. After loading this warning disappears, but because the loading takes quite a while, I am already concerned as a user, if someone hacked my PEN balance

### Expected behavior
My expectation as a user would be, that instead of showing a warning, I could see some info, that the portal is loading.

- [x] Show Skeleton component when Dashboard is loading
- [x] Show no GET PEN when the page is loading


Sharqiewicz:
@pendulum-chain/devs Ready for review. Minor change

ebma (approved):
Nice, looks good to me 👍

PR #551, opened 2024-09-03
Show skeleton on Spacewalk transactions page 

### LINKED ISSUE #513: Transactions page on the portal initially shows empty table without loading spinner
#### Context
When accessing the transactions page on the Portal, an empty table is displayed without a loading spinner. Transactions appear after a few seconds, but initially, it looks like there are none. 

#### Steps to Reproduce
1. Go to https://portal.pendulumchain.org/pendulum/spacewalk/transactions
2. Observe the transactions table.

#### Observed Outcome
- No spinner is displayed.
- Initially, the table is empty.
- After a few seconds, the transactions are displayed.

#### Expected Outcome
- If there are transactions: either see the transactions immediately or see a spinner indicating that the transactions are loading.
- If there are no transactions: display empty transaction table (current view when the wallet is not connected)

ebma:
Shows a skeleton for the Spacewalk transactions table when the data is still loading. 
While the issue ticket asks for a spinner, I think the skeleton is prettier.

Note: There is a little bug that the skeleton does not immediately show when you refresh the page on the Spacewalk transactions view. I'm not sure why exactly and I tried some things to prevent this but it's just a minor issue and unlikely that users will notice as this only pops up when you manually refresh the page on that view. 

Closes #513.


src/pages/spacewalk/transactions/index.tsx
```diff
@@ -108,7 +114,16 @@ function Transactions(): JSX.Element {
 
       return entries;
     };
-    fetchAllEntries().then((res) => setData(res));
+
+    setIsLoading(true);
```
Sharqiewicz, line 118: What do you think about using Tanstack react-query instead?

src/components/Table/index.tsx
```diff
@@ -77,6 +76,20 @@ const Table = <T,>({
 }: TableProps<T>): JSX.Element | null => {
   const totalCount = data.length;
 
+  const showSkeleton = useMemo(() => {
```
Sharqiewicz, line 79: It seems that useMemo might not be necessary in this case. There is no much calculation happening

Sharqiewicz (commented):
@ebma Great changes! 🚀 I noticed one small thing: when I start localhost and am not logged in, the table loads endlessly

src/components/Table/index.tsx
ebma: It's true that there is not much calculation in `showSkeleton` but `showSkeleton` is used by other fields that are memoized which otherwise would rerender all the time. You could argue that neither `showSkeleton` nor `tableData` or `tableColumns` are applying expensive transformations but I still think it should boost performance just a little. Or would you disagree?

src/pages/spacewalk/transactions/index.tsx
ebma: Hmm right, we can do that.

src/components/Table/index.tsx
Sharqiewicz: The performance impact is likely minimal either way, I don’t think it’s crucial to change it. I’m fine with keeping useMemo as it is

src/pages/spacewalk/transactions/index.tsx
Sharqiewicz: Once Implemented, the PR can be merged ✅

ebma: Fixed that 👍

ebma:
@pendulum-chain/product can someone please have a look at the deploy preview and let me know if it's okay the way it works now? Since technically, the transactions need to take the `activeBlockNumber` into account, the table will now show the skeleton each time a new block is coming in. Because for each new block, the table is populated anew. In almost all cases the table will not change though. Please let me know if you think the repeated loading indication is too much and I can change it to only show at the beginning.

annatekl:
> While the issue ticket asks for a spinner, I think the skeleton is prettier.

The table loading time is still quite long

ebma:
Only if you refresh the page on the transactions page or what exactly do you think is long? The time until the skeleton shows or the duration of how long the skeleton stays?

annatekl:
> Only if you refresh the page on the transactions page or what exactly do you think is long? The time until the skeleton shows or the duration of how long the skeleton stays?

the duration the skeleton stays

annatekl:
> That's not what the related issue #513 was about though, was it? 😅 It's about showing a loading indicator and not reducing the loading time.

true, lets do it this way

PR #557, opened 2024-09-12
Add link to subscan explore on the transaction status pop up

### LINKED ISSUE #471: Add link to subscan explore on the transaction status pop-up
## Context
Currently if the swap txn executed by the user is successful or failed on-chain only the status of that is displayed on the pop-up. There is no link to the explorer. 

## Requirment
* Once the user initiates the swap transaction
* If the txn is successful on-chain
    * Add a link to the subscan transaction on the Transaction successful pop-up with text "View on Explorer" above the close button
* If the txn is fails on-chain
    * Add a link to the subscan transaction on the Transaction failed pop-up with text "View on Explorer" above the close button
 
## Note
* This is only for Pendulum
* Reference of the pop-up
image

ebma:
I did not test how it looks for the 'Transaction failed' case as forcing the extrinsic to fail is not easy. 

Note: This currently includes changes that enable Nabla on Pendulum but this is just to facilitate testing with the deploy preview and will be removed before merge. 

Closes #471. 

The button looks like this and it will link to transactions like [this](https://pendulum.subscan.io/extrinsic/0x8d47601c7c634c5b7b7c162539a996cdd2595202631f8d8207b3ec92706dfbcf). 
image


Sharqiewicz:
@ebma Your changes are great ✅ very good job ⭐ 
I also included the fixes requested from @vadaynujra (it is the only opened PR where Forex AMM for Pendulum is enabled)

src/components/nabla/common/TransactionProgress.tsx
```diff
@@ -26,43 +28,61 @@ const getErrorMessage = (data?: ExecuteMessageResult['result']) => {
...
-      <div className="text-center mt-4">
+      <div className="mt-4 text-center">
         <h4 className="text-2xl text-[--text]">
           {mutation.isSuccess ? 'Transaction successful' : 'Transaction failed'}
         </h4>
       </div>
-      {!mutation.isSuccess && !!errorMsg && <p className="text-center mt-1">{errorMsg}</p>}
+      {!mutation.isSuccess && !!errorMsg && <p className="mt-1 text-center">{errorMsg}</p>}
+      <div className="mt-6"></div>
+      {!!explorerUrl && (
```
TorstenStueber, line 79: Just a question because it is the first time I see it that way: Is there a difference between `!!explorerUrl && ...` and `explorerUrl &&` since `explorerUrl` can only be `undefined` or a truthy (nonempty) string? 

[From the official docs](https://react.dev/learn/conditional-rendering#logical-and-operator-):
> React considers `false` as a “hole” in the JSX tree, just like `null` or `undefined`, and doesn’t render anything in its place.

TorstenStueber (approved):
Approved!

A couple of changes just reorder class names. Is this a new prettier or linter fix setting or do local setups between team members deviate or are formatting checks not properly added to CI? 

Would be good to fix this in the config so that we can avoid changes in PRs like this in the future.

prayagd (approved):
LGTM

src/components/nabla/common/TransactionProgress.tsx
ebma: No you are right, there is no need for this here. I just used it to cast it to a boolean value but that's not really necessary here. Removed.

ebma (commented):
I think the reordering of the classNames happens because of [this](https://github.com/pendulum-chain/portal/blob/2c4cd1bb1872c239bb32194f38eefee5e0b3e44e/.prettierrc#L11) plugin that @Sharqiewicz added not too long ago. I locally ran prettier on all files and there were a lot of changes. It's probably best to move this to an extra PR, possibly also adding a precommit hook for running prettier automatically.

ebma:
I followed up to this in https://github.com/pendulum-chain/portal/pull/567.

PR #559, opened 2024-09-17
Improve manual vault selection for spacewalk amplitude and pendulum

### LINKED ISSUE #556: Improve Manual Vault Selection for Spacewalk (Amplitude and Pendulum)
#### Context
Now that we have multiple vaults for the same assets on Spacewalk, there's no clear indication for users that they can bridge larger amounts. The current portal view defaults to one vault (possibly chosen at random), displaying the max issuance based on that specific vault's capacity. This can lead to users with intentions of bridging higher amounts dropping off the flow, mistakenly believing that their request can't be accommodated.

#### Requirements
- [ ] By default, the vault with the highest issuance is selected
- [ ] add a tooltip text that appears when users mouse over the settings button. This text should be "Select Vault".

image (17)
image (18)

ebma:
Changes the automatic vault selection to select the vault with the most issuable tokens (when bridging to Pendulum) or the vault with the most redeemable tokens (when bridging to Stellar) by default. 

Refactors the code a bit to simplify the usage of the `issuableTokens` and `redeemableTokens` on the `ExtendedVault` types. 

Closes #556.

### DIFF
--- a/src/components/Selector/VaultSelector.tsx
+++ b/src/components/Selector/VaultSelector.tsx
@@ -18,9 +18,7 @@ function getMaxTokensForVault(vault: ExtendedRegistryVault, type: 'issue' | 'red
   if (!maxTokens) return '0';
 
   try {
-    const balance: { amount: string } = maxTokens.toJSON();
-
-    return nativeToDecimal(balance.amount.toString()).toFixed(2);
+    return nativeToDecimal(maxTokens).toFixed(2);
   } catch (error) {
     console.error('Error parsing max tokens', error);
     return '0';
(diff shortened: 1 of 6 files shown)


ebma:
@pendulum-chain/product please have a look

vadaynujra:
Thanks for extending the requirement to the 'To Stellar' side as well @ebma. I have tested and can confirm:

1. For each asset, the vault with the highest issuable and redeemable is selected by default - **Expected**
2. Between assets, AUDD is always selected by default - **Not expected**

Do you know if 2 is a result of this change, because from memory it was always the XLM asset (first asset in the dropdown) that was selected by default before?

vadaynujra:
A scenario that I don't necessarily want to test is - when a user bridges an amount with vault A (having the highest issuable), which takes its capacity below vault B, and the user wants to immediately do another bridging transaction. Will the vault selection automatically change to Vault B i.e. the vault with the highest issuable amount?

ebma:
Nice catch with the AUDD asset @vadaynujra. I'm working on a fix. 

> Will the vault selection automatically change to Vault B i.e. the vault with the highest issuable amount?

No it will not, the max issuable and redeemable are at the moment only re-fetched when you e.g. navigate between the Issue/Redeem pages. For the issue case, it's probably not that big of a deal as you as a user would click on the 'I have made the payment' button most of the time and then you are automatically taken to the transactions view. In the redeem case it's different, here we might consider wanting an automatic refresh. If you think this is important let me know.

ebma:
I changed it so that the XLM asset is selected by default again. 
You might notice that the max issuable/redeemable value changes when you navigate back and forth on the pages. This is because every new block authored on our runtimes brings new token prices and thus, the calculation of the issuable/redeemable amounts changes. 

@Sharqiewicz I think the code of `useBridgeSettings` is so intertwined with the context that we might consider just moving everything into the context after all. I also had to move the `extendedVaults` because there were inconsistencies about the max issuable shown on in e.g. the Issue form itself and the vaults in the manual vault selection. Storing them in the context gets rid of this inconsistency.

vadaynujra:
Thanks @ebma, and noted point about the 'issuable' amount changing each block. Anyhting blocking this one from being merged?

vadaynujra:
Let's go ahead without the refresh on the redeem side, as it's very applicable on the issue side.

ebma:
I wanted to wait for a proper review by @Sharqiewicz but if you think this is urgent we can also merge it right away @vadaynujra.

vadaynujra:
Not urgent @ebma and can merge after proper review with @Sharqiewicz . I just wanted to add that the refresh on the redeem side is not blocking this one.

Sharqiewicz (approved):
@ebma Awesome work! ✅

PR #561, opened 2024-09-23
Fix/numericinput edge cases

Sharqiewicz:
- [x] if the user input is "0.23" and the cursor is between "2" and "3" and the user enters "."
expected: ignore the input as there is already one "." in the input field further to the left
actual: input changes to "0.2"
- [x] if the user input is "12.34" and the cursor is between "1" and "2" and the user enters "."
expected: either ignore the input as there is already a "." in the input field further to the right or replace input to "1.234"
actual" input changes to "1.2"


src/components/Form/From/NumericInput/NumericInput.test.tsx
```diff
@@ -153,4 +153,22 @@ describe('NumericInput Component', () => {
...
+    expect(inputElement.value).toBe('0.23');
+  });
+
+  it('Should not cut the number and set properly the new postion of "." in the number', async () => {
+    const { getByPlaceholderText } = render(<NumericInput register={mockRegister} />);
+    const inputElement = getByPlaceholderText('0.0') as HTMLInputElement;
+
+    await userEvent.type(inputElement, '12.34');
+    await userEvent.keyboard('{arrowleft}{arrowleft}{arrowleft}{arrowleft}.');
+    expect(inputElement.value).toBe('1.234');
```
ebma, line 172: This only seems to work in one direction, ie. it only works if the new "." is to the left of the old ".". 
For example, if you enter "12.34" and the cursor is between "3" and "4" and the user enters ".", nothing happens.

Sharqiewicz: I think the desired behaviour is for nothing to happen in this case. I always try Uniswap as an example of how it should work

Sharqiewicz: Do you think we should make it so that the input becomes 123.4? I think that is also a good solution. We could try

ebma: Okay but then I think the logic is inconsistent. When I enter "12.34" on uniswap, and then try to enter a new "." anywhere, I always jump to the end of the input without any change actually happening. But with your current changes, there will be a change but only if I enter the "." to the left of the existing ".". 

So either we don't do anything, or we allow it left _and_ right of the existing ".".

ebma: Fine with me. Though @TorstenStueber s suggestion [here](https://github.com/pendulum-chain/vortex/pull/166#pullrequestreview-2318816396) regarding the paste behaviour sounds good to me.  

>  normalizing all text in handleOnInput (i.e., filter out everything outside [0-9.] and then remove any additional . after the first one and then remove extraneous decimals).

This could be a nice compromise when the user is pasting wrong data, what do you think? Because otherwise, I as a user would be very confused if I wanted to paste an amount (though somehow with invalid formatting) and nothing happens.

Sharqiewicz: @ebma Yes, I think we can do this

TorstenStueber: Yes, I think it better UX to normalize pasted text instead of not to do anything on paste if the text to be pasted is ill-formatted.

TorstenStueber (commented):
This feels very nice, great job.

However, there are cases where the paste event does not behave as expected, namely if I want to paste text within the current text, it will be pasted to the end.

Example:
* input field value is "123"
* my clipboard has the content "4"
* cursor is at the beginning of the input field
* I paste

Expected outcome: 4123 (and cursor between the 4 and 1)
Actual outcome: 1234 (and cursor at the end)

This works correctly on Uniswap.

Sharqiewicz:
@TorstenStueber ✅

ebma (approved):
Nice 👍

src/components/Form/From/NumericInput/helpers.ts
```diff
@@ -40,15 +40,20 @@ export function handleOnKeyPressNumericInput(e: KeyboardEvent): void {
...
+
+  const combinedValue = value.slice(0, selectionStart || 0) + clipboardData + value.slice(selectionEnd || 0);
 
   const [integerPart, ...decimalParts] = combinedValue.split('.');
   const sanitizedValue = integerPart + (decimalParts.length > 0 ? '.' + decimalParts.join('') : '');
 
   e.preventDefault();
   inputElement.value = trimToMaxDecimals(sanitizedValue, maxDecimals);
+
+  const newCursorPosition = (selectionStart || 0) + clipboardData.length;
```
TorstenStueber, line 55: This is already perfect. If you want it to be more than perfect, then you could change this line to 

```suggestion
  const newCursorPosition = (selectionStart || 0) + clipboardData.length - (combinedValue.length - sanitizedValue.length);
```

Reason: the cursor it not at the right position if pasting a valid number with a `.` if there is already a `.` in the number.

TorstenStueber (approved):
Perfection 😍.

src/components/Form/From/NumericInput/helpers.ts
Sharqiewicz: @TorstenStueber Thanks for noticing it, I'll improve it. This input is so battle-tested and has so many test-cases that we can create a npm package named `decimal-input`

TorstenStueber (approved):

PR #563, opened 2024-09-25
Bump polkadot api versions

ebma:
Due to the runtime upgrade to polkadot v1.1.0 on Foucoco, the portal is not able to properly connect to Foucoco anymore. 
When loading the Spacewalk or Staking pages, the following error is thrown
```
index.37e81365.js:120 Uncaught (in promise) RpcError: -32000: Client error: Execution failed: Execution aborted due to trap: wasm trap: wasm `unreachable` instruction executed
WASM backtrace:
error while executing at wasm backtrace:
    0: 0x683a84 - <unknown>!rust_begin_unwind
    1: 0xc2b8 - <unknown>!core::panicking::panic_fmt::h7d22643b0becf577
    2: 0x210e69 - <unknown>!TransactionPaymentApi_query_info
    at checkError (index.37e81365.js:120:411854)
    at RpcCoder.decodeResponse (index.37e81365.js:120:412261)
    at WsProvider.__internal__onSocketMessageResult (index.37e81365.js:120:425224)
    at WebSocket.__internal__onSocketMessage (index.37e81365.js:120:424911)
```

This is because some RPC methods were deprecated and are not available anymore. 

## Solution

We only need to bump the polkadot.js dependency versions. The later versions are aware of the RPC method deprecation and handle the change properly.


src/helpers/substrate.ts
```diff
@@ -1,5 +1,22 @@
 import { EventRecord } from '@polkadot/types/interfaces';
 import { ApiPromise } from '@polkadot/api';
+import Big from 'big.js';
+import { PalletBalancesAccountData } from '@polkadot/types/lookup';
+
+/// This function is used to get the frozen account balance from the account data.
+/// Some types define the frozen balance as `frozen` and others as `miscFrozen` and `feeFrozen`.
```
gianfra-t, line 7: So this is coming from `pallet_balances` types in polkadot. Once you updated the `@polkadot/types` dependency, shouldn't it only be one case?

Or is this such that the portal is compatible with the not-yet-updated versions of the runtime?

src/pages/staking/CollatorsTable.tsx
```diff
@@ -57,7 +59,8 @@ function CollatorsTable() {
         return '0';
       }
       const { data: balance } = await api.query.system.account(walletAccount?.address);
-      return balance.free.sub(balance.frozen).toString();
+      const frozen = getFrozenAccountBalance(balance);
+      return new Big(balance.free.toString() || 0).sub(frozen).toString();
```
gianfra-t, line 63: > balance.free.toString() || 0

Is this extra condition here because `walletAccount` can be `undefined`?

gianfra-t (approved):
Looks good ✅, just some clarification comments.

src/helpers/substrate.ts
```diff
@@ -1,5 +1,22 @@
...
+export function getFrozenAccountBalance(balance: PalletBalancesAccountData) {
+  const { miscFrozen, feeFrozen } = balance;
+  const frozen = balance.get('frozen');
+  if (frozen !== undefined) {
```
Sharqiewicz, line 11: Can we just `if(frozen)`?
Or do we need to check for null, empty string, false or 0?

Sharqiewicz (approved):

src/helpers/substrate.ts
TorstenStueber: I guess it is to make this code useful for all three chains as long as the 1.1.0 is not yet fully rolled out.

TorstenStueber (approved):
👍

ebma:
I was able to simplify this further and get rid of the fee name checks. The problem was that I missed updating the `pendulum-chain/types` package to the latest version as well. Thus, there were multiple  different polkadot-types versions available in the app and the IDE got confused. Building the app failed when just using the `frozen` field as before because the pendulum-types were overriding the polkadot-types. 

Neither our old v0.9.42 runtimes are using `miscFrozen` or `feeFrozen` nor the new v1.1.0 runtime. These fields must have been removed some time ago already though I didn't manage to find out when exactly. 

This is great as it makes the changes of this PR even simpler.

src/hooks/useBuyout/index.ts
```diff
@@ -58,7 +58,7 @@ function handleBuyoutError(error: string) {
 function generateBuyoutExtrinsicPayload(amount: BigNumber) {
   const amountString = amount.toString();
   // We always go the 'buyout' route because the 'exchange' route does not work sometimes
-  return { buyout: { amount: amountString } };
+  return { Buyout: { amount: amountString } };
```
ebma, line 61: This capitalization change is necessary because the updated types now require a type similar to 
```
{ Buyout : any } | { Exchange : any } 
```
and the linting complains if the capitalization doesn't match.

TorstenStueber (approved):
Nice that you could simplify this further as 0.9.42 already used the new types.

By the way: the removal of `src/services/stats/index.tsx` is an unrelated change, correct? You just wanted to clean it up here because you saw that it is not used anywhere?

ebma:
Yes,  `tsc` complained because the stats page was still using `miscFrozen` and `feeFrozen`. Since we don't use it anyway, I removed it. IIRC it was used some time ago to expose token-api data with a Netlify function but we replaced this already a long time ago with our dedicated token-api.

Issue #474, opened 2024-05-29
Rename the first Swap pool UI column and add tool tip


prayagd:
## Context
Currently, the few columns of the Swap pool UI are not very clear, this ticket is to update them in a way that it is simpler to understand for the majority of DeFi / AMM users with knowledge of 2-asset pool AMMs. 

## Requirement
* Rename the first column of the swap UI to "Asset"
* Add a tooltip next to "Pool liabilities" and "Reserves" titles
* Tool tip message for Pool liabilities - Amount of liquidity provided initially by LPs
* Tool tip message for Reserves - Amount of tokens in the pool

## Note
* This is both for the Pendulum and Foucoco deployment

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/portal/474?planning-poker) with Zenhub @ebma @Sharqiewicz

ebma:
> Instead of showing name's of the asset, show tickers of the assets for ex. instead of Stellar BRL show BRL.s

A note for the implementation of this. Instead of showing the value of the 'name' field, we should then instead show the 'symbol' field. For the current set of deployed contracts this means that 'Mock BRL' will show as 'BRL'. @prayagd if you want this to say 'BRL.s' instead, we would need to redeploy the contracts on Foucoco yet again.

prayagd:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/portal/474?planning-poker) with Zenhub @TorstenStueber

TorstenStueber:
@prayagd it is not correct to rename "Pool liabilities" to "Deposits". What even is a deposit? The number of LP tokens in circulation or the number of pool tokens deposited into the pool. The latter is the "reserves".

If you intend to express the number of LP tokens in circulation: That isn't the "total liabilities" either, instead it is the "total supply", a different value managed by the pool. Total liabilities and total supply are similar but will deviate further and further over time.

prayagd:
@TorstenStueber Fair point, should we instead add tooltips to these titles? so the end user understands this term.
What is the definition Pool liabilites? as i consider Reserves to be Number of tokens in the pool.

TorstenStueber:
[...]

PR #565, opened 2024-09-26
show Nabla Swap tooltips, change column names

Sharqiewicz:
🦈 PR
---
- [x] Rename the first column of the swap UI to "Asset"
- [x] Add a tool tip next to "Pool liabilities" and "Reserves" titles
- [x] Tool tip message for Pool liabilities - Amount of liquidity provided initially by LPs
- [x] Tool tip message for Reserves - Amount of tokens in the pool


TorstenStueber:
Sorry @Sharqiewicz this ticket was not ready for dev and should not have been in the Ready column. There is still ongoing discussion.

@prayagd The definition of Pool liabilities as "Amount of liquidity provided initially by LPs" is wrong (see also my comment [here](https://github.com/pendulum-chain/portal/issues/474#issuecomment-2283321794)).

Pool liabilities deviate from the "Amount of liquidity provided initially by LPs" further and further due to
* slippage
* accumulated fees

Eventually the liability will deviate a lot from the "Amount of liquidity provided initially by LPs".

There is no simple, intuitive way to explain what "liability" means. As I wrote before, it is a number used for internal accounting to determine the current slippage.

vadaynujra:
@prayagd not sure if you saw my comment here, but shall we keep the column names consistent in adding the word `Pool`, or not? Let's remove it from both (like Nabla have) as the user is already in the Swap Pools page?

prayagd:
> @prayagd not sure if you saw my comment https://github.com/pendulum-chain/portal/issues/474#issuecomment-2383612899, but shall we keep the column names consistent in adding the word Pool, or not? Let's remove it from both (like Nabla have) as the user is already in the Swap Pools page?

Agree to the suggestion

Sharqiewicz:
@vadaynujra @prayagd What is the conclusion?

prayagd:
@Sharqiewicz I would suggest getting rid of the tool tips and just renaming the first column to "Asset", Suggesting this because the ticket has been going on for a longer time and all further changes can be done in a separate ticket. do you agree @TorstenStueber @ebma ?

ebma:
Sure, we can reduce the scope here and follow up in another ticket.

TorstenStueber:
Yes, agree. Let's get this ticket done with a reduced scope and let's open another ticket if we consider this necessary. I am currently writing the Forex AMM docs in a way that it should work well with the current table layout.

prayagd:
@Sharqiewicz let's reduce the scope and just rename the first column to "Asset"

Sharqiewicz:
@prayagd Is it good now?

ebma (approved):
Looks good to me. 

This PR also addresses https://github.com/pendulum-chain/portal/issues/592 now so we can close that ticket as well afterwards.

TorstenStueber (approved):

PR #569, opened 2024-09-30
Enable Forex AMM on Pendulum

Sharqiewicz:
⭐ New menu option available: ForexAMM ⭐


TorstenStueber (approved):
👍

prayagd:
- Currently on swap screen after entering a number in the input field it shows the price of the asset, Please change it to the amount based on user's input  for example user enters 0.5 for DOT and the price is $4.04, then it should be 0.5*4.04 = $2.02, it should be done for both fields. Please let me know if not clear. 

image

prayagd:
@pendulum-chain/devs The Deposit in the backstop pool is not working, can you please check
image

ebma:
> Update the "Powered by Vortex finance" hyperlink to vortexfinance.co

@prayagd shouldn't it rather be 'Update the "Powered by Vortex finance" text to "Powered by Nabla finance"'? It's not powered by vortex but by Nabla so the link is correct.

vadaynujra:
@ebma following agreement of terms with Nabla team, we are now allowed to call it powered by Vortex Finannce and therefore also link it to the Vortex Finance website.

ebma:
Okay but it doesn't make sense, does it? It would need to be called "Powering Vortex finance" if anything. It has nothing to do with Vortex?

TorstenStueber:
@prayagd what swap direction did you use? USDC to EURC or DOT?

TorstenStueber:
@prayagd I checked the problem with the backstop pool. There are two issues: 1) we don't properly show an error message and 2) the error actually happening here is that the value of each backstop pool token in this configuration is negative and therefore it is not possible to charge the backstop pool.

The value of a pool token is calculated as the sum of
* the amount tokens in the backstop pool (which is 0)
* for each swap pool: the "excess liquidity" of the swap pool price converted to the backstop pool token

The excess liquidity of each swap pool is
* EURC: -1.159, convered to USD: -1.28 USD
* DOT: 0.2, converted to USD: 0.83 USD
* USDC.axl: 0
* USDC: 0

So the total worth is -1.28 + 0.83 = -0.45 USD is negative.

## Why did it happen
This can usually only happen initially if the backstop pool does not contain any significant amount of tokens yet and there was a trade and after that a price shift.

## How to resolve
To resolve just transfer, e.g., 1 EURC to the address of the EURC swap pool (`6g2pcZdxWKQjF36Mb5ebEDAx2NRw2zSYv82Vwa93yz3mzbbR`), no deposit, just normal transfer. After that you can charge the backstop pool.

prayagd:
> @prayagd what swap direction did you use? USDC to EURC or DOT?

DOT to EURC.s

prayagd:
@TorstenStueber Thank, the transfer worked

Sharqiewicz:
> Expanding the details section on swap screen does not work, it just increases the font size. I have attached a screenshot on how it should look like
> Update the "Powered by Vortex finance" hyperlink to vortexfinance.co
> Add icon's before the name of the asset on the swap pool column

@prayagd Implemented 🦈

> Currently on swap screen after entering a number in the input field it shows the price of the asset, Please change it to the amount based on user's input for example user enters 0.5 for DOT and the price is $4.04, then it should be 0.5*4.04 = $2.02, it should be done for both fields. Please let me know if not clear.

In Progress 🦈

Sharqiewicz:
@prayagd Ready 🦈🟢 Please review 🥺

src/components/nabla/Swap/To.tsx
```diff
@@ -74,14 +74,25 @@ export function To({
...
         </Button>
       </div>
-      <div className="mt-1 flex items-center justify-between text-neutral-500 dark:text-neutral-300">
-        <div className="mt-px text-sm">{toToken ? <NablaTokenPrice address={toToken.id} fallback="$ -" /> : '$ -'}</div>
+      <div className="flex items-center justify-between mt-1 text-neutral-500 dark:text-neutral-300">
+        <div className="mt-px text-sm">
+          {toToken ? (
+            <NablaTokenPrice
+              formatByAmount={true}
+              currentTokenAmount={Number(toAmountQuote.data?.amountOut.approximateStrings.atLeast4Decimals) || 0}
```
TorstenStueber, line 88: Let's generally avoid `number` for financial applications and only use the `Big.js` library or native `bigint`.

```suggestion
              currentTokenAmount={Big(toAmountQuote.data?.amountOut.approximateStrings.atLeast4Decimals ?? "0")}
```

src/components/nabla/common/NablaTokenPrice.tsx
```diff
@@ -5,14 +5,33 @@ export type TokenPriceProps = {
...
+}: TokenPriceProps): JSX.Element | null {
   const { data, isLoading } = useNablaTokenPrice(address);
   if (isLoading) return <NumberLoader />;
 
   if (data === undefined) return fallback;
 
+  if (formatByAmount) {
+    const currentAmountPrice = (Number(data.approximateStrings.atLeast2Decimals) * (currentTokenAmount || 0)).toFixed(
+      2,
+    );
```
TorstenStueber, line 27: Let's generally avoid `number` for financial applications and only use the `Big.js` library or native `bigint`.

```suggestion
    const currentAmountPrice = (data.preciseBigDecimal.mul(currentTokenAmount ?? 0)).toFixed(
      2, 0
    );
```

src/components/nabla/common/NablaTokenPrice.tsx
```diff
@@ -5,14 +5,33 @@ export type TokenPriceProps = {
   address: string;
   prefix?: ReactNode;
   fallback?: ReactNode;
+  currentTokenAmount?: number;
```
TorstenStueber, line 8: Let's generally avoid `number` for financial applications and only use the `Big.js` library or native `bigint`.

```suggestion
  currentTokenAmount?: Big;
```

src/components/nabla/Swap/From.tsx
```diff
@@ -54,15 +54,24 @@ export function From<FormFieldValues extends FieldValues, TFieldName extends Fie
...
         </Button>
       </div>
-      <div className="mt-1 flex items-center justify-between text-neutral-500 dark:text-neutral-400">
+      <div className="flex items-center justify-between mt-1 text-neutral-500 dark:text-neutral-400">
         <div className="mt-px text-sm">
-          {fromToken ? <NablaTokenPrice address={fromToken.id} fallback="$ -" /> : '$ -'}
+          {fromToken ? (
+            <NablaTokenPrice
+              formatByAmount={true}
+              currentTokenAmount={form.watch(fromFormFieldName)}
```
TorstenStueber, line 68: `fromFormFieldName` is of type string. With my suggestions the type of `currentTokenAmount` should be `Big` (see my comments below).

Instead of reparsing the input string, it might be better to forward the `fromAmount` in the `Swap` component to the `From` component as the parsing is then already robustly done in `useSwapComponent`.

TorstenStueber (changes requested):
I have a few comments regarding the use of `number` instead of big number formats. It's just good practice that we use big numbers as some of the assets can have many decimals (e.g., 18) and precision will be lost.

vadaynujra:
@Sharqiewicz I just realised we don't show the limit / cap anywhere to the user. Is that something we should add in?

vadaynujra:
Reference from Nabla 
image

Sharqiewicz:
@TorstenStueber Do you think we should add "Cap" to the "Liabilities" column?

TorstenStueber:
@Sharqiewicz @vadaynujra I would generally think that we should have these conversations within the ticket before someone picks it up. Nevertheless, what is the definition of the Cap?

src/components/nabla/common/NablaTokenPrice.tsx
```diff
@@ -1,30 +1,30 @@
...
   const { data, isLoading } = useNablaTokenPrice(address);
   if (isLoading) return <NumberLoader />;
 
   if (data === undefined) return fallback;
 
   if (formatByAmount) {
-    const currentAmountPrice = (Number(data.approximateStrings.atLeast2Decimals) * (currentTokenAmount || 0)).toFixed(
-      2,
-    );
+    const currentAmountPrice = Big(data.approximateStrings.atLeast2Decimals).mul(currentTokenAmount).toFixed(2);
```
TorstenStueber, line 26: Please write `toFixed(2, 0)`, otherwise it rounds to nearest instead of rounding down.

vadaynujra:
@TorstenStueber limits on swap and bsp.

TorstenStueber:
Let's move this discussion to the ticket where we already discuss table captions: https://github.com/pendulum-chain/portal/issues/474

prayagd (approved):
Confirming that all the bugs that were raised are now fixed and working

TorstenStueber (approved):

PR #577, opened 2024-10-01
configure dependabot

Sharqiewicz:
🦈 PR - in progress
---

Dependabot will check daily for new versions of dependencies but will only apply updates for security vulnerabilities. Non-security updates are ignored to reduce noise. The packages will be updated to the latest versions [?].

Pull requests created by Dependabot will be prefixed with “bot-chore” in their commit messages, limited to a maximum of 5 open PRs. Each PR will be automatically labeled with “dependencies” and “security,” assigned to @Sharqiewicz . All updates will be grouped into a single PR for easier management.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,6 +17,3 @@ updates:
       - "Sharqiewicz"
     reviewers:
       - "Sharqiewicz"
-    groups:
-      security-dependencies:
-        applies-to: security-updates
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,7 @@ updates:
     schedule:
       interval: "daily"
     commit-message:
-      prefix: "fix"
+      prefix: "bot-chore:"
     open-pull-requests-limit: 10
     ignore:
       - dependency-name: "*"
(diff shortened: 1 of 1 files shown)


Sharqiewicz:
@ebma @TorstenStueber ready for quick review

.github/dependabot.yml
```diff
@@ -0,0 +1,26 @@
...
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    commit-message:
+      prefix: "bot-chore:"
+    open-pull-requests-limit: 10
+    ignore:
+      - dependency-name: "*"
```
TorstenStueber, line 11: Doesn't this mean that we ignore all dependencies?

.github/dependabot.yml
```diff
@@ -0,0 +1,26 @@
...
+    security-updates: true
```
TorstenStueber, line 12: Couldn't find this top level setting in the [dependabot docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file), sure that it is correct?

TorstenStueber (commented):
The PR looks generally good, I have two comments, though.

.github/dependabot.yml
```diff
@@ -0,0 +1,26 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    commit-message:
+      prefix: "bot-chore:"
+    open-pull-requests-limit: 10
```
ebma, line 9: Can we limit this to 5 instead? 10 is a lot 👀

ebma (approved):

PR #594, opened 2024-10-14
Show the portal pen balance only for transferrable tokens

### LINKED ISSUE #485: Show the Portal PEN balance only for transferrable tokens
## Context
Currently the balance shown in the top right corner for the portal UI, adds up all categories of token balances for ex. locked etc etc. This causes confusion because the transferrable balance could be different. 

## Requirement
* Show only the transferrable tokens as balance in the top right corner

ebma:
Makes the button in the top-right hand corner always show the transferable balance instead of the total balance.

Closes #485.


Sharqiewicz (approved):
LGTM

prayagd (approved):
Tested both Pendulum and Amplitude only shows transferrable balance, good to go from my end

PR #606, opened 2024-10-25
remove unnecessary useEffect in useConnectWallet

Sharqiewicz:
I simplified the code by removing unnecessary state management, as the wallets are only filtered once and don’t require reactivity


ebma (approved):
Cool 👍 

This might help me with https://github.com/pendulum-chain/portal/pull/595 so I'll merge this right away.

PR #613, opened 2024-10-31
Use favicon.ico instead of favicon.png

Sharqiewicz:
I don't know why we had two versions of favicon-  `.png` and `.ico`. I removed the png one and connected .ico as it is the standard


Sharqiewicz:
⚠️  After the migration, routing stopped working because the generation of the `_redirects` file was accidentally removed. Without this file, Netlify couldn’t properly redirect all pages to index.html, causing broken links for client-side routes. This PR includes a hotfix to restore the _redirects file generation, ensuring that Netlify correctly handles routing for that application

Issue #609, opened 2024-10-28
Incorrect balance display 0


Sharqiewicz:
![Screenshot 2024-10-28 at 15 27 27](https://github.com/user-attachments/assets/14818b6b-6e7a-474f-a5e8-9819d57fc9d3)

When balance is `0`, the component shows `-0`

PR #614, opened 2024-10-31
dont calculate TransferableBalance when balance is 0

Sharqiewicz:
🦈 PR 🟢
---

### Issue description
After recent changes, we encountered an issue where accounts with a zero balance resulted in a negative transferable balance, specifically -0.00 (See image in the #609).

### The cause
For accounts with a free balance of 0, `calculateTransferableBalance` can produce a negative value (specifically -0.001) because the EXISTENTIAL_DEPOSIT is set to 0.001, causing a deduction even when the free balance is zero.


ebma (approved):
Nice catch 👍

PR #626, opened 2024-12-13
Added +PEN to APR

prayagd:
Add PEN to APR

### Context
This is for the liquidity campaign participants that they would receive PEN tokens as rewards along with the APR


ebma (approved):
Works. Next time consider putting more context in the description of the PR as the necessity of this change wouldn't be clear to me if @vadaynujra hadn't told me about it before. If we ever wonder why we added this change in the first place, it would be nice if we could go back to this PR and get the reminder we need. 

Feel free to hit the merge button @prayagd.

prayagd:
@ebma Makes sense added context now, hope that makes it clear for future references too

PR #632, opened 2025-01-09
Fix buggy calculation of transferable amount for security deposit

### LINKED ISSUE #631: BUG: Fix calculation of security deposit for EURC bridging
### Context
Observed that teh portal is wrongly calculating the PEN required in the user's account to pay for deposit, while the portal shows the right PEN amount.
image

### To do 
Find reason why this behavior and add fix

ebma:
The problem occurs only for accounts with a large balance of PEN tokens. The `useAccountBalance` hook returned the values with _pretty formatting_. This means that large amounts were depicted with commas, e.g. `10,000.00`.  The consumer of this balance could wrongly interpret the number as 10 PEN (if the comma is expected to denote the decimals) which is exactly what happened when transforming the transferable balance for calculating the security deposit. 

This PR changes the return type of `useAccountBalance` to return numbers instead. 

Closes #631.

### DIFF
--- a/src/components/Wallet/modals/DisconnectModal/index.tsx
+++ b/src/components/Wallet/modals/DisconnectModal/index.tsx
@@ -67,7 +67,7 @@ const WalletDropdownMenu = ({
       />
     </div>
     <p className="my-6 truncate text-center text-2xl font-bold" title={`${balance} ${tokenSymbol}`}>
-      {balance} {tokenSymbol}
+      {balance && prettyNumbers(balance)} {tokenSymbol}
     </p>
     <Button className="bg-base-300" size="sm" onClick={removeWalletAccount}>
       <ArrowLeftEndOnRectangleIcon className="mr-2 w-5" />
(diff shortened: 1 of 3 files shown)


vadaynujra:
Not getting shown the error anymore, but the security deposit calculation isn't being shown anymore @ebma 
image

ebma:
Prices are only available on the production deployment due to CORS restrictions. So you can't test it on the deploy preview.

src/components/Wallet/modals/DisconnectModal/index.tsx
```diff
@@ -65,11 +66,11 @@ const WalletDropdownMenu = ({
         inline={true}
       />
     </div>
-    <p className="my-6 text-2xl font-bold text-center truncate" title={`${balance} ${tokenSymbol}`}>
+    <p className="my-6 truncate text-center text-2xl font-bold" title={`${balance} ${tokenSymbol}`}>
       {balance} {tokenSymbol}
```
Sharqiewicz, line 70: Don't we want `{balance && prettyNumbers(balance)} {tokenSymbol}` here as well?

Sharqiewicz (approved):

src/components/Wallet/modals/DisconnectModal/index.tsx
ebma: True, changed 👍

PR #633, opened 2025-01-10
Fix crash when address is copied to clipboard.

gianfra-t:
### What:

A bug was detected that would crash the application when `CopyablePublicKey` was used (and clicked). Root cause was an undetermined bug in `react-toastify`, when `toast` was called.

### How:

Simply updating the dependency solved the issue.

### DIFF
--- a/src/shared/showToast.ts
+++ b/src/shared/showToast.ts
@@ -23,59 +23,59 @@ const ToastProperties: Record<ToastMessage, ToastSettings> = {
     message: 'Could not establish connection with the bridge...',
     options: {
       toastId: ToastMessage.BRIDGE_CONNECTION_ERROR,
-      type: toast.TYPE.ERROR,
+      type: 'error',
     },
   },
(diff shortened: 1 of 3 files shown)


src/shared/showToast.ts
```diff
@@ -23,67 +23,67 @@ const ToastProperties: Record<ToastMessage, ToastSettings> = {
...
-      type: toast.TYPE.ERROR,
+      type: 'error',
     },
   },
   [ToastMessage.NODE_CONNECTION_ERROR]: {
     message: 'Error while connecting to the node. Refresh the page to re-connect.',
     options: {
       toastId: ToastMessage.NODE_CONNECTION_ERROR,
-      type: toast.TYPE.ERROR,
+      type: 'error',
```
gianfra-t, line 33: Property `TYPE` is not there anymore at this version. I do not think it is too bad to use a string since `options` is typed.

Sharqiewicz (approved):

PR #636, opened 2025-01-28
[READY] Use raw html elements with daisyui css instead of react-daisyui

Sharqiewicz:
## The problem:
Dropdown doesn't work on mobile browsers.
`react-daisyui` package elements seem to not properly handle the mobile browser events.


ebma (approved):
Seems to work overall 👍  I just noticed that the manual vault selector dialog is now scrollable if you enable the check box for manual selection. It seems to me like it's scrollable because of the elements in the dropdown. I assume if there were more than just two elements for selection, the dialog would allow scrolling down even further. Can you confirm that the dialog is now scrollable @Sharqiewicz?

Sharqiewicz:
@ebma Good catch! Vault Selector is using `react-daisyui` implementation and it doesn't work on mobile. I need to fix this one too. ✅

Sharqiewicz:
@ebma The dropdowns are now refactored ✅

ebma (approved):
Amazing, let's go 🙏

PR #642, opened 2025-03-04
add both formats to copyable address component

gianfra-t:


Sharqiewicz (changes requested):
By my mind to not complicate the situation in the app, we should go for new addresses only and show an announcement at the top of the website that there is a new address format and place a link to an article explaining everything. wdyt @gianfra-t @ebma

Sharqiewicz:
@pendulum-chain/devs

gianfra-t:
Fine by me! Then the change should happen automatically in this repo.

ebma:
> We should go for new addresses only and show an announcement at the top of the website that there is a new address format and place a link to an article explaining everything.

I agree. It's a nice idea with opening the extra dialog with two formats on click, however it's also not ideal in the sense that if I as a user am confused about seeing a different address, I surely wouldn't expect to see a dialog open when I click on a copyable address. So this does not really address the confusion users are seeing as it's still kind of hidden. 

Let's just add the banner and a link to the documentation shared with us.

